PR #172, opened 2020-12-11
feat: generate docs for jsii submodules

### LINKED ISSUE #1: jsii-docgen does not generate markdown for the submodules
If a typescript project contains an exported module (submodule for jsii), no information is generated in the markdown when jsii-docgen is ran against it.

Chriscbr:
Fixes #1 

- Adds docs for all classes, structs, interfaces, and enums in submodules
- Changes class initialization examples to include submodule paths, e.g. `new foo.MyClass(...)` instead of `new MyClass(...)`
- Updates class docs to include a "submodule" section if the class has an associated namespace
- Modifies tsconfig to allow using flatMap

I tested this with docs for [projen](https://github.com/projen/projen), and the would-be updated docs are here: https://gist.github.com/Chriscbr/41f7f7e7d1d358dc6ebf0f8cb277b937

### DIFF
--- a/.projenrc.js
+++ b/.projenrc.js
@@ -25,8 +25,18 @@ const project = new TypeScriptLibraryProject({
   },
   releaseToNpm: true,
   projenUpgradeSecret: 'PROJEN_GITHUB_TOKEN',
+  tsconfig: {
+    compilerOptions: {
+      lib: ['es2018', 'es2019.array'],
+    },
+  },
+  jestOptions: {
+    typescriptConfig: {
+      compilerOptions: {
+        lib: ['es2018', 'es2019.array'],
+      },
+    },
+  },
 });
-
-project.tsconfig.compilerOptions.lib = ['es2018', 'es2019.array'];
 
 project.synth();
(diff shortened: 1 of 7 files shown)


.projenrc.js
```diff
@@ -25,6 +25,18 @@ const project = new TypeScriptLibraryProject({
   },
   releaseToNpm: true,
   projenUpgradeSecret: 'PROJEN_GITHUB_TOKEN',
+  tsconfig: {
```
eladb, line 28: Add a comment why this is needed

.projenrc.js
```diff
@@ -25,6 +25,18 @@ const project = new TypeScriptLibraryProject({
...
+    compilerOptions: {
+      lib: ['es2018', 'es2019.array'],
+    },
+  },
+  jestOptions: {
+    typescriptConfig: {
+      compilerOptions: {
+        lib: ['es2018', 'es2019.array'],
```
eladb, line 36: Reuse compiler options

eladb (changes requested):
The names in the TOC tables should also include the submodule prefix.

For example, in projen, the `Github` class should be named `github.Github`

eladb (approved):

PR #356, opened 2021-07-01
feat: Enhanced API Reference

iliapolo:
Introduce the new API reference that was developed as part of the work on [ConstructHub](https://github.com/cdklabs/construct-hub).

This PR is 99% a straight up copy of the code from the [`docgen`](https://github.com/cdklabs/construct-hub-webapp/tree/main/src/api/docgen) api in the [`construct-hub-webapp`](https://github.com/cdklabs/construct-hub-webapp) project.

Example of how this change manifests in the `API.md` of construct hub itself: 

https://github.com/cdklabs/construct-hub/blob/epolon/testing-nextgen-docgen/API.md

BREAKING CHANGE: The tool no longer accepts any arguments and operates solely on the `.jsii` assembly produced during package build. In addition, the `API.md` file went through a major overhaul, so we consider this a breaking change as well.


src/cli.ts
```diff
@@ -1,22 +1,52 @@
...
 
+  const assembly = await createAssembly();
 
-  if (args._.length === 0) {
-    const output = args.output ?? 'API.md';
-    await renderSinglePageModule(process.cwd(), output);
-  } else {
-    const output = args.output ?? 'dist';
-    await renderFiles(args._ as string[], output);
+  const docs = new Documentation({
```
iliapolo, line 17: Since generating the API reference is slightly more involved now, requiring dependant assemblies, and operating on a single package, I removed support for passing any input to the CLI.

I think the main use-case for this tool is to generate the `API.md` file as part of the normal jsii project build procedure, so this is what it currently supports.

iliapolo:
Still in draft while I try to sort out link handling in way that both supports construct-hub and also doesn't break in standard markdown viewers.

src/docgen/render/markdown.ts
```diff
@@ -0,0 +1,209 @@
...
+      // passed that, markdown just renders `#` signs.
+      // lets see if and when we'll hit this limit.
+      throw new Error('Unable to render markdown. Header limit (6) reached.');
+    }
+
+    const content: string[] = [];
+    if (this.header) {
+      const anchor = sanitize(this.id ?? '');
+      const heading = `${'#'.repeat(headerSize)} ${this.header}`;
+      const headerSpan = !!process.env.HEADER_SPAN;
```
iliapolo, line 167: This is nasty, i'm aware. 

Its just an escape hatch so that produce working links by default, but also support producing the links that `construct-hub` currently [relies](https://github.com/cdklabs/construct-hub-webapp/blob/main/src/components/PackageDocs/PackageDocs.tsx#L86) on.

This will be gone soon.

src/docgen/render/markdown.ts
```diff
@@ -0,0 +1,214 @@
...
+
+    const content: string[] = [];
+    if (this.header) {
+      const anchor = sanitize(this.id ?? '');
+      const heading = `${'#'.repeat(headerSize)} ${this.header}`;
+
+      // This is nasty, i'm aware.
+      // Its just an escape hatch so that produce working links by default, but also support producing the links that construct-hub currently relies on.
+      // This will be gone soon.
+      // Note though that cross links (i.e links dependencies will not work yet regardless)
```
iliapolo, line 171: For reviewers: Please note this comment

RomainMuller (approved):

eladb (approved):
This is too awesome

PR #359, opened 2021-07-05
feat: transliterate user defined content

iliapolo:
Use `jsii-rosetta` to transliterate assemblies before constructing the reflect TypeSystem.

This makes it so README code snippets as well as `@example` code blocks for inline documentation are also transliterated, providing a full language specific page. 

> Currently still in draft because it uses an unpublished version of `@jsii/spec` and `jsii-rosetta` pending a jsii release.

### Implementation Notes

Previously, the `Documentation` class accepted a **required** `reflect.Assembly` property. This meant that the creation of the type-system was the responsibility of the caller. This API is somewhat cumbersome,  and now that the assemblies need to be transliterated beforehand, this is even more so. 

The PR changes the input to the `Documentation` class, so that it only requires local paths that contain assembly files, it then proceeds to transliterate those assemblies and create the type-system on its own. 

Since the creation of the type-system is an `async` process, the `Documentation` class offers a `static async` function that needs to be awaited on, instead of directly using the constructor, which is now private. 

So the usage now becomes:

```ts
const docs = await Documentation.forLocalPackage(rootPath, assembliesPath);
const page = docs.render().render();
```

In addition, a new functionality of generating documentation for remote packages was added:

```ts
const docs = await Documentation.forRemotePackage('@aws-cdk/aws-ecr', '1.110.1');
const page = docs.render().render();
```

BREAKING CHANGE: The `Documentation` class can no longer be directly instantiated, use static factory methods instead.


src/docgen/render/markdown.ts
```diff
@@ -170,7 +170,6 @@ export class Markdown {
       // This will be gone soon.
       // Note though that cross links (i.e links dependencies will not work yet regardless)
       const headerSpan = !!process.env.HEADER_SPAN;
-      console.log(`Header span: ${headerSpan}`);
```
iliapolo, line 173: Leftover from a previous PR... :\

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
-          `Generating documentation for ${options.language} is not supported`,
-        );
-    }
 
-    if (options.submoduleName) {
-      this.submodule = this.findSubmodule(options.submoduleName);
+  /**
+   * Create a `Documentation` object for a remote package available in the npm registry.
+   */
+  public static async forRemotePackage(name: string, version: string, options?: DocumentationOptions): Promise<Documentation> {
```
eladb, line 90: Maybe this should be called `fromNpm()` and `fromDirectory()` ("remote" is not that clear).

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+    return withTempDir(async (workdir: string) => {
+      const env = {
+        ...process.env,
+        // npm fails with EROFS if $HOME is read-only, even if it won't write there
+        HOME: os.tmpdir(),
+      };
+
+      await exec('npm install npm@7', {
```
eladb, line 98: add a comment explaining why you need v7

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+        cwd: workdir,
+        env,
+      });
+      await exec(`${workdir}/node_modules/.bin/npm install --ignore-scripts --no-bin-links --no-save ${name}@${version}`, {
```
eladb, line 102: Add a comment why `--ignore-scripts` is important from a security standpoint (and add a test to protect against regression)

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+      });
+
+      return Documentation.forLocalPackage(path.join(workdir, 'node_modules', name), { ...options, assembliesDir: workdir } );
+    });
+  }
+
+  /**
+   * Create a `Documentation` object for a package available in the local file system.
+   */
+  public static async forLocalPackage(root: string, options?: FromLocalPackageDocumentationOptions): Promise<Documentation> {
```
eladb, line 114:
```suggestion
  public static async forLocalPackage(root: string, options: FromLocalPackageDocumentationOptions = {}): Promise<Documentation> {
```

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+
+      let transpile, language;
+
+      switch (options?.language ?? 'ts') {
+        case 'python':
+          language = TargetLanguage.PYTHON;
+          transpile = new PythonTranspile();
+          break;
+        case 'ts':
+          transpile = new TypeScriptTranspile();
```
eladb, line 146: `language = TYPESCRIPT`?

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+  private constructor(
+    private readonly assembly: reflect.Assembly,
+    private readonly transpile: Transpile) {
   }
 
   /**
    * Generate markdown.
    */
-  public render(): Markdown {
+  public render(options?: RenderOptions): Markdown {
```
eladb, line 165:
```suggestion
  public renderMarkdown(options: RenderOptions = {}): Markdown {
```

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+  } finally {
+    process.chdir(cwd);
+    await fs.remove(workdir);
+  }
+}
+
+async function createAssembly(name: string, tsDir: string, loose: boolean, language?: TargetLanguage): Promise<reflect.Assembly> {
+  const ts = new reflect.TypeSystem();
+  for (let dotJsii of await glob.promise(`${tsDir}/**/.jsii`)) {
+    if (language) {
```
eladb, line 219: TypeScript should be skipped probably?

eladb (commented):
Looks great

README.md
```diff
@@ -16,17 +16,8 @@ As a library:
...
-await ts.loadFile('my/project/.jsii');
-
-const docs = new Documentation({
-  assembly: ts.findAssembly('my-project-name'),
-  language: 'ts',
-  readme: false,
-});
 
+const docs = await Documentation.forLocalPackage('.', { language: 'ts' });
 const markdown = docs.render().render(); // returns a markdown string
```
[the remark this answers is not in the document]
RomainMuller: That or

```suggestion
const markdown = docs.renderer().renderMarkdown(); // returns a markdown string
```

src/docgen/view/documentation.ts
RomainMuller: There are also interesting questions about how the package is fetched... It might be relevant to rename this `forRegistryPackage`, especially if this will transparently use system/user-wide configuration for which registry is being used...

src/docgen/transpile/transpile.ts
```diff
@@ -374,7 +374,7 @@ export interface TranspiledModuleLike {
  */
 export interface Transpile {
   /**
-   * The language of the tranpiler.
+   * The language of the transpiler.
    */
   language: string;
```
RomainMuller, line 379: Should this property be readonly? (Although unrelated to the present change)

```suggestion
  readonly language: string;
```

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+export interface DocumentationOptions {
+
+  /**
+   * Which language to generate docs for.
    *
-   * @default true
+   * @default 'ts'
    */
-  readonly apiReference?: boolean;
+  readonly language?: string;
```
RomainMuller, line 57: I'd be hard-pressed determining what are valid values to pass here... Should I pass `python` or `py` here? Is it `chsarp`, `C#`, `.NET`, or `dotnet`?

Either this needs to be a union, or an `enum`, or the documentation needs to tell me how I should name those...

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+ */
+export interface FromLocalPackageDocumentationOptions extends DocumentationOptions {
 
   /**
-   * Generate documentation only for a specific submodule.
+   * A local directory containing jsii assembly files that will
+   * comprise the type-system.
    *
-   * @default - Documentation is generated for all submodules.
+   * @default - the root package directory will be used.
```
RomainMuller, line 77: What does this mean? I'm not sure I get it...

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+      const env = {
+        ...process.env,
+        // npm fails with EROFS if $HOME is read-only, even if it won't write there
+        HOME: os.tmpdir(),
+      };
+
+      await exec('npm install npm@7', {
+        cwd: workdir,
+        env,
+      });
```
RomainMuller, line 101: I tend to prefer using the form with separate arguments, which can avoid arguments parsing surprises... (Although this particular case is rather safe).

You need to enable `shell` or else this may not work well in Windows...

```suggestion
      await exec('npm', ['install', 'npm@7'], {
        cwd: workdir,
        env,
        shell: true,
      });
```

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+      await exec(`${workdir}/node_modules/.bin/npm install --ignore-scripts --no-bin-links --no-save ${name}@${version}`, {
+        cwd: workdir,
+        env,
+      });
```
RomainMuller, line 105: In this particular instance, the version with separated arguments is safer, too... And this also requires `shell` to be Windows friendly (not sure `.bin/npm` will effectively exist in Windows... it may be `npm.exe` or `npm.bat` or something... and `shell: true` is how you ensure Windows isn't stupid about those).

```suggestion
      await exec(
        `${workdir}/node_modules/.bin/npm`,
         [
           'install',
           '--ignore-scripts',
           '--no-bin-links',
           '--no-save',
           `${name}@${version}`,
         ], {
        cwd: workdir,
        env,
        shell: true,
      });
```

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
     }
 
-    this.includeApiReference = options.apiReference ?? true;
-    this.includeReadme = options.readme ?? true;
+    // normally the assemblies are located in subdirectories
+    // of the root package dir (i.e ./node_modules)
+    const assembliesDir = options?.assembliesDir ?? root;
+
+    const manifest = JSON.parse(await fs.readFile(manifestPath, 'utf-8'));
+    return Documentation.forAssembly(manifest.name, assembliesDir);
```
RomainMuller, line 125: This is nitpick/style/optional... I have grown to like unstructuring assignments when I only care about a few properties of a possibly large object... It makes it "obvious" what I'm interested in up-front, and also allows `node` to manage memory a little better (no need to retain the whole object in memory!)

```suggestion
    const { name } = JSON.parse(await fs.readFile(manifestPath, 'utf-8'));
    return Documentation.forAssembly(name, assembliesDir);
```

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+      await fs.copy(assembliesDir, workdir);
+
+      let transpile, language;
+
+      switch (options?.language ?? 'ts') {
+        case 'python':
+          language = TargetLanguage.PYTHON;
+          transpile = new PythonTranspile();
+          break;
+        case 'ts':
```
RomainMuller, line 145: As I said above... this is inconsistent... Why `ts` and not `typescript` (or why `python` and not `py`). Since this is not type-checked in the current form, it could probably use some alternate names/syntaxes to avoid surprising folks.

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+          transpile = new TypeScriptTranspile();
+          break;
+        default:
+          throw new Error(`Unsupported language: ${options?.language}`);
```
RomainMuller, line 149: I like it when such errors actually also tell you what **is** supported, so you're not left wondering.

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
+      const assembly = await createAssembly(assemblyName, workdir, options?.loose ?? true, language);
+      return new Documentation(assembly, transpile);
+    });
+
+  }
+
+  private constructor(
+    private readonly assembly: reflect.Assembly,
+    private readonly transpile: Transpile) {
   }
```
RomainMuller, line 160: Style nitpick (this would reduce unrelated changes if one was to ever add more arguments to this constructor)

```suggestion
  private constructor(
    private readonly assembly: reflect.Assembly,
    private readonly transpile: Transpile,
  ) {
  }
```

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
 
   /**
-   * Lookup a submodule by a submodule name given by the caller.
+   * Lookup a submodule by a submodule name.
    */
-  private findSubmodule(submoduleName: string): reflect.Submodule {
-    const submodules = this.assembly.submodules.filter(
-      (s) => s.name === submoduleName,
+  private findSubmodule(assembly: reflect.Assembly, submodule?: string): reflect.Submodule {
```
RomainMuller, line 185: It makes no sense to me that the `submodule` parameter of a `findSubmodule` method is optional... How do you go and find a submodule by name if you're not given its name?

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
-      throw new Error(`Submodule ${submoduleName} not found`);
+      throw new Error(`Submodule ${submodule} not found`);
     }
 
     if (submodules.length > 1) {
-      throw new Error(`Found multiple submodules with name: ${submoduleName}`);
+      throw new Error(`Found multiple submodules with name: ${submodule}`);
     }
 
     return submodules[0];
```
RomainMuller, line 198: Optional -- I like to write this kind of stuff as (not using `suggestion` here because GitHub won't allow me in this particular context):

```ts
    const [submodule, ...rest] = assembly.submodules.filter((s) => s.name === submodule);
    if (submodule == null) {
      throw new Error(`Submodule named "${submodule}" does not exist in "${assembly.name}@${assembly.version}"`);
    }
    if (rest.length > 0) {
      throw new Error(`Found multiple submodules named "${submodule}" in "${assembly.name}@${assembly.version}"`);
    }
    return submodule
```

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,227 @@
...
   }
 }
+
+async function withTempDir<T>(work: (workdir: string) => Promise<T>): Promise<T> {
+  const workdir = await fs.mkdtemp(path.join(os.tmpdir(), path.sep));
```
RomainMuller, line 203:
```suggestion
  const workdir = await fs.mkdtemp(path.join(os.tmpdir(), ''));
```

test/docgen/view/documentation.test.ts
```diff
@@ -1,48 +1,45 @@
...
 import { Documentation } from '../../../src';
-import { Assemblies } from '../assemblies';
 
-const assembly: reflect.Assembly = Assemblies.instance.withoutSubmodules;
-const assemblyWithSubmodules: reflect.Assembly = Assemblies.instance.withSubmodules;
+const ASSEMBLIES = `${__dirname}/../../__fixtures__/assemblies`;
+
+// this is a little concerning...we should be mindful
+// if need to keep increasing this.
+jest.setTimeout(2 * 60 * 1000);
```
RomainMuller, line 7: My experience is that this can be avoided by reducing `jest` concurrency (which by default is somewhat unbounded, and causes excessive IO-queuing, especially in CI environments with constrained resources).

README.md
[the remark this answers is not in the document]
iliapolo: So the `docs.render` method actually returns a `Markdown` class, so it should probably be:

```ts
const markdown = docs.toMarkdown().render()
```

Or something. I'm gonna leave it as is for now and once @MrArnoldPalmer merges the JSON rendering we'll refactor it to make sense for both formats.

src/docgen/view/documentation.ts
iliapolo: I like `forRegistryPackage` 👍  - later on we can add a registry URL option for it so that we won't have to use system-wide configurations.

src/docgen/view/documentation.ts
iliapolo: I'd like to...but `TypScript` isn't actually a member of the [`TargetLanguage`](https://github.com/aws/jsii/blob/main/packages/jsii-rosetta/lib/languages/target-language.ts) enum. I didn't want to venture into adding it right now, not even sure it makes sense.

src/docgen/view/documentation.ts
iliapolo: Like I mentioned we will address this confusion later on

src/docgen/view/documentation.ts
iliapolo: It is skipped, this is exactly the scenario where the target language is undefined, again stemmed from the fact typescript isn't really a target language in rosetta.

src/docgen/view/documentation.ts
iliapolo: Done

src/docgen/view/documentation.ts
iliapolo: Done

src/docgen/transpile/transpile.ts
iliapolo: Done

src/docgen/view/documentation.ts
iliapolo: Yeah, its kind of tricky...

This refers to the required argument in the `forLocalPackage` function. It means that all assemblies are expected to be available in the root directory (normally `<package-dir>/node_modules).

You feel this needs further explanation here? it's one of those properties that I hope don't get used much, and if they are, than its because the user is looking for this support (like we do in `forRemotePackage`).

src/docgen/view/documentation.ts
iliapolo: `exec` doesn't actually have this form, I guess you're thinking of `spawn`? Which doesn't have support for `promisify`. In any case I added an `async` implementation of `spawn` (taken from your code in the transliterator :)).

So...done :)

src/docgen/view/documentation.ts
iliapolo: Done

src/docgen/view/documentation.ts
iliapolo: 👍

src/docgen/view/documentation.ts
iliapolo: Yeap, this was a leftover, I actually check for `null` before i invoke this method, so not needed

src/docgen/view/documentation.ts
iliapolo: 👍

src/docgen/view/documentation.ts
iliapolo: I'll stick to my style here for now, but I have taken the more detailed error message suggestion :)

src/docgen/view/documentation.ts
iliapolo: Why? This will actually break because it will produce directories like `/tmp123456`, as opposed to `/tmp/123456.

test/docgen/view/documentation.test.ts
iliapolo: Ok, good to know - we'll keep an eye on this.

src/docgen/view/documentation.ts
iliapolo: Done

src/docgen/view/documentation.ts
eladb: Nit: I think the "from" convention for factories is more robust. "For" works for documentation but not for other things. 

For example, "documentation for directory" doesn't work while "Documentation [object] from directory" works.

But this is just me being me

src/docgen/view/documentation.ts
iliapolo: Changed to enum 👍

src/docgen/view/documentation.ts
iliapolo: Yeap, changed to `enum`

src/docgen/view/documentation.ts
iliapolo: @RomainMuller I ended up using an enum like class instead to add a transformation from a literal string to the object, and spare this switch casing from the caller.

src/docgen/view/documentation.ts
iliapolo: Done

iliapolo: @eladb Since I didn't want to rely on a published npm package which has lifecycle hooks, I needed to run `forRegistryPackage` on a local package I can control. This got me thinking that `forRegistryPackage` should actually be `forPackage` (or `fromPackage` as you would have liked it :) ) and a user can simply pass any npm installable package, be it a registry identifier or a path to a local tarball. 

So that's what I did and that is what I use in the test. 

https://github.com/cdklabs/jsii-docgen/blob/4ec588d5ecf0b139b38220af80f7d0296807f345/test/docgen/view/documentation.test.ts#L56

src/docgen/view/documentation.ts
iliapolo: I agree that `forDirectory` doesn't work, but I also think its because of the "directory" - it doesn't have any semantic meaning. So the solution here could also be to come up with a proper semantic name for it. 

It is now `forProject` - which signals that its meant to be used on locally built projects with a `package.json` in its root directory.

iliapolo:
@eladb @RomainMuller This is ready now for merge following jsii latest [release](https://github.com/aws/jsii/releases/tag/v1.31.0)

src/docgen/transpile/transpile.ts
```diff
@@ -1,5 +1,48 @@
 import * as reflect from 'jsii-reflect';
 
+/**
+ * Supported languages to generate documentation in.
+ */
+export class Language {
```
eladb, line 6: Curious - what's the reason this is not a simple enum?

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,310 @@
...
+      if (await fs.pathExists(target) && !options.name) {
+        throw new Error('\'options.name\' must be provided when installing local packages.');
+      }
+
+      const name = options?.name ?? extractPackageName(target);
+
+      const env = {
+        ...process.env,
+        // npm fails with EROFS if $HOME is read-only, even if it won't write there
+        HOME: os.tmpdir(),
```
eladb, line 113: feels like this should be something below `tmpdir()` no (like `mkdtmp()`)

src/docgen/view/documentation.ts
```diff
@@ -1,121 +1,310 @@
...
+      // npm7 is needed so that we also install peerDependencies - they are needed to construct
+      // the full type system.
+      await spawn('npm', ['install', 'npm@7'], {
+        cwd: workdir,
+        env,
+        shell: true,
+        stdio: ['ignore', 'inherit', 'inherit'],
+      });
+
+      console.log('Installing package');
```
eladb, line 125:
```suggestion
      console.log(`Installing package ${target}`);
```

src/docgen/view/documentation.ts
eladb: I usually do: `await fs.mkdtemp(path.join(os.tmpdir(), 'jsii-docgen.'));`

eladb (approved):

Issue #369, opened 2021-07-14
Python generation fails on some packages


iliapolo:
For example:

```ts
test('@aws-cdk/aws-route53-patterns', async () => {
  const docs = await Documentation.forPackage('@aws-cdk/aws-route53-patterns@1.106.0', {
    language: Language.PYTHON,
  });
  const markdown = docs.render({ linkFormatter: (t: TranspiledType) => `#custom-${t.fqn}` });
  expect(markdown.render()).toMatchSnapshot();
});
```

```console
Debug Failure. False expression: Containing file should be a module.

      at checkImportMetaProperty (node_modules/typescript/lib/typescript.js:59527:22)
      at checkMetaProperty (node_modules/typescript/lib/typescript.js:59502:24)
      at checkExpressionWorker (node_modules/typescript/lib/typescript.js:61722:28)
      at checkExpression (node_modules/typescript/lib/typescript.js:61614:38)
      at checkNonNullExpression (node_modules/typescript/lib/typescript.js:57015:37)
      at checkPropertyAccessExpression (node_modules/typescript/lib/typescript.js:57060:85)
      at checkExpressionWorker (node_modules/typescript/lib/typescript.js:61693:28)
      at checkExpression (node_modules/typescript/lib/typescript.js:61614:38)
      at maybeCheckExpression (node_modules/typescript/lib/typescript.js:60827:34)
      at checkBinaryExpression (node_modules/typescript/lib/typescript.js:60795:25)
      at checkExpressionWorker (node_modules/typescript/lib/typescript.js:61734:28)
      at checkExpression (node_modules/typescript/lib/typescript.js:61614:38)
      at checkExpressionStatement (node_modules/typescript/lib/typescript.js:64083:13)
      at checkSourceElementWorker (node_modules/typescript/lib/typescript.js:66611:28)
      at checkSourceElement (node_modules/typescript/lib/typescript.js:66489:17)
      at Object.forEach (node_modules/typescript/lib/typescript.js:317:30)
      at checkSourceFileWorker (node_modules/typescript/lib/typescript.js:66826:20)
      at checkSourceFile (node_modules/typescript/lib/typescript.js:66794:13)
      at getDiagnosticsWorker (node_modules/typescript/lib/typescript.js:66882:17)
      at getDiagnostics (node_modules/typescript/lib/typescript.js:66868:24)
      at Object.getEmitResolver (node_modules/typescript/lib/typescript.js:35702:13)
[...]

PR #370, opened 2021-07-20
feat: loose mode to ignore all errors

iliapolo:
Currently the `loose` property is being used to signal acceptable failures to `jsii-rosetta` during the transliteration process.
Apparently there are [failures](https://github.com/cdklabs/jsii-docgen/issues/369) which are not covered by setting `loose: true` in rosetta.

In order to be able to generate API reference even for those types of failures, we broaden the meaning of `loose` in the context of `jsii-docgen` such that any transliteration failure will be caught and silently ignored.


test/docgen/view/documentation.test.ts
```diff
@@ -10,7 +10,7 @@ const LIBRARIES = `${__dirname}/../../__fixtures__/libraries`;
 
 // this is a little concerning...we should be mindful
 // if need to keep increasing this.
-jest.setTimeout(2 * 60 * 1000);
+jest.setTimeout(3 * 60 * 1000);
```
iliapolo, line 13: The new test I added takes more than 2 minutes... :\

eladb: Ouch

RomainMuller (approved):

eladb (approved):

PR #373, opened 2021-07-22
fix: rosetta failures causes fallback to typescript

iliapolo:
This PR reverts a previous misguided decision to [ignore](https://github.com/cdklabs/jsii-docgen/pull/370) rosetta failures and fallback to typescript assemblies. 

We decided we never want to show typescript code in the documentation of other languages, a better experience is actually to fail and either introduce more heuristics in rosetta to bypass such failures, or make the necessary adjustments to the published package. 

In addition, we used to transliterate the entire type-system (i.e all dependent assemblies) and not just the top level assembly. 
The rational being that code snippets might come from those assemblies when expanding arguments for python docs. 

Problem is that this means that a transliteration failure in a deeply nested dependency, that most likely doesn't have any affect on the documentation, prevents package transliteration. This can act as a sort of poison pill because many packages depend on the same core libraries. 

Also, we aren't currently even rendering those code snippets in the docs, so there is no good reason to do it.

### DIFF
--- a/test/docgen/view/documentation.test.ts
+++ b/test/docgen/view/documentation.test.ts
@@ -10,7 +10,7 @@ const LIBRARIES = `${__dirname}/../../__fixtures__/libraries`;
 
 // this is a little concerning...we should be mindful
 // if need to keep increasing this.
-jest.setTimeout(3 * 60 * 1000);
+jest.setTimeout(60 * 1000);
 
 
 describe('extractPackageName', () => {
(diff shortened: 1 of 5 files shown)


test/docgen/view/documentation.test.ts
```diff
@@ -10,7 +10,7 @@ const LIBRARIES = `${__dirname}/../../__fixtures__/libraries`;
 
 // this is a little concerning...we should be mindful
 // if need to keep increasing this.
-jest.setTimeout(3 * 60 * 1000);
+jest.setTimeout(60 * 1000);
```
iliapolo, line 13: Bonus!

RomainMuller (approved):

PR #383, opened 2021-07-29
chore: migrate mergify behavior to actions

NGL321:
Replaces mergify functionality that cannot be replicated by Github settings with actions.

- Semantic PR failure comment
- Core contributor label


iliapolo:
@NGL321 This looks good in general. 

The problem is that since enabling auto-merge on a PR is a manual action, it means our automatic dependency upgrade PR's won't get automatically merged. Meaning we (the on call most likely) would have to go to each of those PR's and click a bunch of buttons - which is exactly what we set out to prevent with this mechanism. 

I think this problem needs to be addressed before we can disable mergify and move over. Perhaps using `GraphQL` github API's to enable auto-merge on those PR's as part of their approval. 

> https://github.community/t/enabling-auto-merge-on-a-pr-using-the-api/150797/5

eladb (changes requested):
Furthermore - this repo is managed by projen so all chances should go through projen and ideally at the base library and across all our projects.

Please work with @iliapolo to follow up.

PR #388, opened 2021-08-02
fix: install does not include dev-deps

RomainMuller:
Also, switches to using `fork`, as this provides improved start-up time
when the child process is `node`, and use a dependency-provided `npm@7`
instead of installing one on each execution.

### DIFF
--- a/src/docgen/view/documentation.ts
+++ b/src/docgen/view/documentation.ts
@@ -6,6 +6,7 @@ import * as glob from 'glob-promise';
 import * as reflect from 'jsii-reflect';
 import { TargetLanguage } from 'jsii-rosetta';
 import { transliterateAssembly } from 'jsii-rosetta/lib/commands/transliterate';
+import { bin as npmBin } from 'npm/package.json';
 import { Markdown } from '../render/markdown';
 import { PythonTranspile } from '../transpile/python';
 import { Transpile, Language, TranspiledType } from '../transpile/transpile';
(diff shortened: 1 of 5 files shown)


src/docgen/view/documentation.ts
```diff
@@ -6,6 +6,7 @@ import * as glob from 'glob-promise';
 import * as reflect from 'jsii-reflect';
 import { TargetLanguage } from 'jsii-rosetta';
 import { transliterateAssembly } from 'jsii-rosetta/lib/commands/transliterate';
+import { bin as npmBin } from 'npm/package.json';
```
eladb, line 9: So we are bundling npm into our lambda bundles. Is that the intent?

src/docgen/view/documentation.ts
```diff
@@ -114,29 +115,28 @@ export class Documentation {
...
         'install',
         // this is critical from a security perspective to prevent
         // code execution as part of the install command using npm hooks. (e.g postInstall)
         '--ignore-scripts',
         // ensures npm does not insert anything in $PATH
         '--no-bin-links',
+        // Ignore a class of dependency conflicts
+        '--force',
+        // Include dev & peer dependencies,
+        '--include', 'dev',
```
eladb, line 129: Why are we including dev deps? We shouldn't need to

RomainMuller: Almost none of the CDK documentation transliterates cleanly without it's dev dependencies... Think about example code that uses `devDependencies`... to demonstrate proper use.

src/docgen/view/documentation.ts
RomainMuller: That would be. However I don't actually think this can work with `esbuild` bundling, as the `bin/npmcli.js` file wouldn't be where it's supposed to be, I think... It'd need to be handled as external...

eladb: So I guess that should be configured somehow, no?

src/docgen/view/documentation.ts
eladb: I see. I was thinking this is for generating API references but I understand why this is needed for sample code transliteration.

eladb:
@RomainMuller is this pr still relevant?

eladb:
@RomainMuller Abandon?

Chriscbr:
Looks like this auto-closed when I renamed the default branch from master to main. Feel free to re-open this or create a new PR.

PR #431, opened 2021-09-07
chore: set author to aws

Chriscbr:


package.json
```diff
@@ -29,8 +29,8 @@
     "projen": "npx projen"
   },
   "author": {
-    "name": "Elad Ben-Israel",
-    "email": "benisrae@amazon.com",
+    "name": "Amazon Web Services",
+    "url": "https://aws.amazon.com",
     "organization": false
```
nija-at, line 34: I believe this should be set to `true`.

nija-at (changes requested):

nija-at (approved):

PR #435, opened 2021-09-11
feat: add code examples for properties

Chriscbr:
Add code examples for all properties within "Properties" sections. This does not add code examples for parameters (e.g. for fields in constructors or methods).

Even though the code examples include the types of the properties, we cannot remove the "Type: xxx" labels below them because they may include _links_ to other types, and code snippets cannot include links in GitHub markdown. (We can probably work around this for the Construct Hub by using the JSON format instead).

Resolves https://github.com/cdklabs/jsii-docgen/issues/436

### DIFF
--- a/src/docgen/transpile/java.ts
+++ b/src/docgen/transpile/java.ts
@@ -8,6 +8,11 @@ const toCamelCase = (text?: string) => {
   return Case.camel(text ?? '');
 };
 
+const toUpperCamelCase = (test?: string) => {
+  const camelCase = toCamelCase(test);
+  return camelCase.charAt(0).toUpperCase() + camelCase.slice(1);
+};
+
 // [1, 2, 3] -> [[], [1], [1, 2], [1, 2, 3]]
 const prefixArrays = <T>(arr: T[]): T[][] => {
   const out: T[][] = [[]];
@@ -166,20 +171,24 @@ export class JavaTranspile extends transpile.TranspileBase {
   public parameter(
     parameter: reflect.Parameter,
   ): transpile.TranspiledParameter {
+    const typeRef = this.typeReference(parameter.type);
     return {
       name: parameter.name,
       parentType: this.type(parameter.parentType),
-      typeReference: this.typeReference(parameter.type),
+      typeReference: typeRef,
       optional: parameter.optional,
+      signatureOrGetter: this.formatProperty(parameter.name, typeRef),
     };
   }
 
   public property(property: reflect.Property): transpile.TranspiledProperty {
+    const typeRef = this.typeReference(property.type);
     return {
       name: property.name,
       parentType: this.type(property.parentType),
-      typeReference: this.typeReference(property.type),
+      typeReference: typeRef,
       optional: property.optional,
+      signatureOrGetter: this.formatProperty(property.name, typeRef),
     };
   }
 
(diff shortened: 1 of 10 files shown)


src/docgen/view/property.ts
```diff
@@ -38,6 +38,8 @@ export class Property {
       md.lines('');
     }
 
+    md.code(this.transpile.language.toString(), this.transpiled.signatureOrGetter);
```
iliapolo, line 41: Since constants currently delegate to properties, this applies to constants as well:

https://gist.github.com/Chriscbr/4de488d36ece0d124149ffdb70688230#cfn_resource_type_name-

It shouldn't

src/docgen/transpile/typescript.ts
```diff
@@ -209,4 +213,14 @@ export class TypeScriptTranspile extends transpile.TranspileBase {
...
   }
+
+  private formatProperty(
+    name: string,
+    typeReference: transpile.TranspiledTypeReference,
+  ): string {
+    const tf = typeReference.toString({
+      typeFormatter: (t) => t.name,
+    });
+    return `public readonly ${name}: ${tf}`;
```
iliapolo, line 224:
```suggestion
    return `public readonly ${name}: ${tf};`;
```

src/docgen/transpile/transpile.ts
```diff
@@ -144,6 +144,11 @@ export interface TranspiledParameter {
    * Whether or not the parameter is optional.
    */
   readonly optional: boolean;
+  /**
+   * The signature of the property, or its getter if the language
+   * supports that.
+   */
+  readonly signatureOrGetter: string;
```
iliapolo, line 151:
```suggestion
  readonly declaration: string;
```

src/docgen/transpile/java.ts
```diff
@@ -373,4 +382,18 @@ export class JavaTranspile extends transpile.TranspileBase {
...
+
+  private formatProperty(
+    name: string,
+    typeReference: transpile.TranspiledTypeReference,
+  ): string {
+    const tf = typeReference.toString({
+      typeFormatter: (t) => t.name,
+    });
+    if (tf.includes(' OR ')) {
+      return `public java.lang.Object get${toUpperCamelCase(name)}()`;
```
iliapolo, line 394:
```suggestion
      return `public java.lang.Object get${toUpperCamelCase(name)}();`;
```

src/docgen/transpile/java.ts
```diff
@@ -373,4 +382,18 @@ export class JavaTranspile extends transpile.TranspileBase {
...
+    } else {
+      return `public ${tf} get${toUpperCamelCase(name)}()`;
```
iliapolo, line 396:
```suggestion
      return `public ${tf} get${toUpperCamelCase(name)}();`;
```

iliapolo (changes requested):

src/docgen/view/property.ts
Chriscbr: Good callout.

src/docgen/transpile/typescript.ts
Chriscbr: Done.

src/docgen/view/property.ts
```diff
@@ -5,9 +5,10 @@ import { Transpile, TranspiledProperty, TranspiledType } from '../transpile/tran
 export class Property {
   private readonly transpiled: TranspiledProperty;
   constructor(
-    transpile: Transpile,
+    private readonly transpile: Transpile,
     private readonly property: reflect.Property,
     private readonly linkFormatter: (type: TranspiledType) => string,
+    private readonly isConstant: boolean = false,
```
iliapolo, line 11: No need. You can just use `property.const`.

iliapolo (changes requested):

iliapolo (approved):

PR #456, opened 2021-10-14
feat: c# support

MrArnoldPalmer:
Adds support for `c#` documentation transpilation. Adds
`/src/docgen/transpile/csharp.ts` with formatters for classes, structs,
interfaces, and enums for `c#`.

### DIFF
--- a/.projen/tasks.json
+++ b/.projen/tasks.json
@@ -49,22 +49,10 @@
         }
       ]
     },
-    "test:compile": {
-      "name": "test:compile",
-      "description": "compiles the test code",
-      "steps": [
-        {
-          "exec": "tsc --noEmit --project tsconfig.dev.json"
-        }
-      ]
-    },
     "test": {
       "name": "test",
       "description": "Run tests",
       "steps": [
-        {
-          "spawn": "test:compile"
-        },
         {
           "exec": "jest --passWithNoTests --all --updateSnapshot"
         },
--- a/package.json
+++ b/package.json
@@ -11,7 +11,6 @@
   "scripts": {
     "clobber": "npx projen clobber",
     "compile": "npx projen compile",
-    "test:compile": "npx projen test:compile",
     "test": "npx projen test",
     "build": "npx projen build",
     "test:watch": "npx projen test:watch",
@@ -47,7 +46,7 @@
     "jest-junit": "^12",
     "json-schema": "^0.3.0",
     "npm-check-updates": "^11",
-    "projen": "^0.30.3",
+    "projen": "^0.31.0",
     "standard-version": "^9",
     "ts-jest": "^26.5.6",
     "typescript": "~4.4.4"
--- a/yarn.lock
+++ b/yarn.lock
@@ -702,9 +702,9 @@
     "@types/node" "*"
 
 "@types/glob@*":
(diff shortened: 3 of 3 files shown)


src/docgen/transpile/csharp.ts
```diff
@@ -0,0 +1,226 @@
...
+    const signatures = [`public ${name}(${params})`];
+    const invocations = [isInitializer
+      ? `new ${name}(${params});`
+      : `${type.name}.${name}(${params});`];
+
+    return {
+      name,
+      parentType: type,
+      import: this.formatImport(type),
+      parameters: callable.parameters,
```
Chriscbr, line 74: Would prefer to sort the parameters so that the docs look consistent between languages.

iliapolo: Yeah, all parameters should be sorted by optionality:

https://github.com/cdklabs/jsii-docgen/blob/58a61628ea9b3c951b1a88304198e5e094f5f8a8/src/docgen/transpile/java.ts#L117

test/docgen/view/__snapshots__/struct.test.ts.snap
```diff
@@ -1,5 +1,83 @@
...
+# Initializer <a name=\\"[object Object].Initializer\\"></a>
+
+\`\`\`csharp
+using Amazon.CDK.AWS.ECR;
+
+new CfnPublicRepositoryProps {
+    repositoryCatalogData
+    repositoryName
+    repositoryPolicyText
+    tags
```
iliapolo, line 19: With all other languages we came up with a way to show the type of the property inlined here, as well as an indication of optionality, showing the required ones first. Is there something we can do here?

src/docgen/transpile/csharp.ts
```diff
@@ -0,0 +1,226 @@
...
+      typeReference: typeRef,
+      optional: parameter.optional,
+      declaration: this.formatProperty(parameter.name, typeRef),
+    };
+  }
+
+  public property(property: reflect.Property): transpile.TranspiledProperty {
+    const typeRef = this.typeReference(property.type);
+    return {
+      name: property.name,
```
iliapolo, line 126: Shouldn't this be `PascalCased`? Basing it of on https://github.com/aws-samples/aws-cdk-examples/blob/master/csharp/application-load-balancer/src/ApplicationLoadBalancer/ApplicationLoadBalancerStack.cs

test/docgen/view/__snapshots__/property.test.ts.snap
```diff
@@ -1,5 +1,22 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
+exports[`csharp snapshot 1`] = `
+" \`repositoryCatalogData\`<sup>Optional</sup> <a name=\\"Amazon.CDK.AWS.ECR.CfnPublicRepositoryProps.property.repositoryCatalogData\\"></a>
+
+\`\`\`csharp
+object repositoryCatalogData
```
iliapolo, line 7: I wonder if there's something mode idiomatic here. Our AWS Dotnet docs show a getter like so:

```chsharp
public virtual string AppId { get; }
```

Is this appropriate here?

PR #481, opened 2021-11-01
feat: add `toJson` method to `Documentation`

MrArnoldPalmer:
Allows rendering language specific documentation to a JSON object in
addition to a markdown string. This allows the consumer more flexibility
to render the structure of the API reference as needed without
regenerating markdown documents.

Additionally removes the `linkFormatter` argument from the `ApiReference`
and type views constructors and moves it to be passed to the `toMarkdown`
method instead. This makes sense because link formatting is markdown
specific and is not needed for the markdown rendering. Additionally, a similar
feature may be useful on the json side in the future to format nested type strings.

BREAKING CHANGE: renames `Documentation.render` method to
`Documentation.toMarkdown`.

### DIFF
--- a/src/docgen/schema.ts
+++ b/src/docgen/schema.ts
@@ -1,0 +1,335 @@
+/**
+ * Describes a type.
+ */
+export interface TypeSchema {
+
+  /**
+   * The type FQN. If missing, the type is a reference (array, map..)
+   */
+  readonly fqn?: string;
+
+  /**
+   * The name of the type (Map, List...)
+   */
+  readonly name: string;
+
+  /**
+   * The various types of the reference.
+   */
+  readonly types?: TypeSchema[];
+}
+
+/**
+ * Describes a property.
+ */
+export interface PropertySchema {
+
+  /**
+   * Unique id of the property.
+   */
+  readonly id: string;
+
+  /**
+   * Name.
+   */
+  readonly name: string;
+
+  /**
+   * Whether or not the property is optional.
+   */
+  readonly optional: boolean;
+
+  /**
+   * Whether or not the property is deprecated.
+   */
+  readonly deprecated: boolean;
+
+  /**
+   * Deprecation reason (if applicable)
+   */
+  readonly deprecationReason?: string;
+
+  /**
+   * Doc strings of the property.
+   */
+  readonly docs: string;
+
+  /**
+   * The type of the property.
+   */
+  readonly type: TypeSchema;
+
+  /**
+   * The default value of the property
+   */
+  readonly default?: string;
+}
+
+/**
+ * Describes a parameter.
+ */
+export interface ParameterSchema extends PropertySchema {}
+
+/**
+ * Common properties of a method.
+ */
+export interface MethodSchemaBase {
+
+  /**
+   * Unique id of the method.
+   */
+  readonly id: string;
+
+  /**
+   * Code snippet to display.
+   */
+  readonly snippet: string;
+
+  /**
+   * Method parameters.
+   */
+  readonly parameters: ParameterSchema[];
+}
+
+/**
+ * Describes a constructor.
+ */
+export interface InitializerSchema extends MethodSchemaBase {}
+
+/**
+ * Describes a method.
+ */
+export interface MethodSchema extends MethodSchemaBase {
+
+  /**
+   * Method name.
+   */
+  readonly name: string;
+}
+
+/**
+ * Describes a class.
+ */
+export interface ClassSchema {
+
+  /**
+   * Unique class id.
+   */
+  readonly id: string;
+
+  /**
+   * Class name.
+   */
+  readonly name: string;
+
+  /**
+   * Interfaces this class implements.
+   */
+  readonly interfaces: TypeSchema[];
+
+  /**
+   * Doc string for the class.
+   */
+  readonly docs: string;
+
+  /**
+   * Class initializer.
+   */
+  readonly initializer?: InitializerSchema;
+
+  /**
+   * Instance methods.
+   */
+  readonly instanceMethods: MethodSchema[];
+
+  /**
+   * Static functions.
+   */
+  readonly staticFunctions: MethodSchema[];
+
+  /**
+   * Properties.
+   */
+  readonly properties: PropertySchema[];
+
+  /**
+   * Constants.
+   */
+  readonly constants: PropertySchema[];
+}
+
+/**
+ * Describes a construct.
+ */
+export interface ConstructSchema extends ClassSchema {}
+
+/**
+ * Describes a struct.
+ */
+export interface StructSchema {
+
+  /**
+   * Unique struct id.
+   */
+  readonly id: string;
+
+  /**
+   * Struct name.
+   */
+  readonly name: string;
+
+  /**
+   * Doc string of the struct.
+   */
+  readonly docs?: string;
+
+  /**
+   * How to initialize the struct.
+   */
+  readonly initializer: string;
+
+  /**
+   * Properties.
+   */
+  readonly properties: PropertySchema[];
+}
+
+/**
+ * Describes an interface.
+ */
+export interface InterfaceSchema {
+
+  /**
+   * Unique interface id.
+   */
+  readonly id: string;
+
+  /**
+   * Interface name.
+   */
+  readonly name: string;
+
+  /**
+   * Interfaces that this interface extends.
+   */
+  readonly interfaces: TypeSchema[];
+
+  /**
+   * Types implementing this interface.
+   */
+  readonly implementations: TypeSchema[];
+
+  /**
+   * Doc string for this interface.
+   */
+  readonly docs: string;
+
+  /**
+   * Methods.
+   */
+  readonly instanceMethods: MethodSchema[];
+
+  /**
+   * Properties.
+   */
+  readonly properties: PropertySchema[];
+}
+
+/**
+ * Describes an enum member.
+ */
+export interface EnumMemberSchema {
+
+  /**
+   * Unique enum member id.
+   */
+  readonly id: string;
+
+  /**
+   * Member name.
+   */
+  readonly name: string;
+
+  /**
+   * Whether or not the member is deprecated.
+   */
+  readonly deprecated: boolean;
+
+  /**
+   * Deprecation reason (if applicable).
+   */
+  readonly deprecationReason?: string;
+
+  /**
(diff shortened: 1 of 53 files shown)


eladb:
Can you share an example of a json file produced from a real world module?

src/docgen/schema.ts
```diff
@@ -0,0 +1,335 @@
...
+
+/**
+ * Describes an enum member.
+ */
+export interface EnumMemberSchema {
+
+  /**
+   * Unique enum member id.
+   */
+  readonly id: string;
```
eladb, line 246: Why not call this fqn?

src/docgen/schema.ts
```diff
@@ -0,0 +1,335 @@
...
+
+  /**
+   * Member name.
+   */
+  readonly name: string;
+
+  /**
+   * Whether or not the member is deprecated.
+   */
+  readonly deprecated: boolean;
```
eladb, line 256:
```suggestion
  readonly deprecated?: boolean;
```

src/docgen/schema.ts
```diff
@@ -0,0 +1,335 @@
...
+
+  /**
+   * Doc string for this interface.
+   */
+  readonly docs: string;
+
+  /**
+   * Methods.
+   */
+  readonly instanceMethods: MethodSchema[];
```
eladb, line 230:
```suggestion
  readonly methods: MethodSchema[];
```

src/docgen/schema.ts
```diff
@@ -0,0 +1,335 @@
...
+
+  /**
+   * Interface name.
+   */
+  readonly name: string;
+
+  /**
+   * Interfaces that this interface extends.
+   */
+  readonly interfaces: TypeSchema[];
```
eladb, line 215:
```suggestion
  readonly bases: TypeSchema[];
```

src/docgen/schema.ts
```diff
@@ -0,0 +1,335 @@
...
+
+  /**
+   * Types implementing this interface.
+   */
+  readonly implementations: TypeSchema[];
+
+  /**
+   * Doc string for this interface.
+   */
+  readonly docs: string;
```
eladb, line 225:
```suggestion
  readonly docs?: string;
```

src/docgen/schema.ts
```diff
@@ -0,0 +1,335 @@
...
+ */
+export interface StructSchema {
+
+  /**
+   * Unique struct id.
+   */
+  readonly id: string;
+
+  /**
+   * Struct name.
```
eladb, line 177: ?

src/docgen/schema.ts
```diff
@@ -0,0 +1,335 @@
...
+
+  /**
+   * Doc string of the struct.
+   */
+  readonly docs?: string;
+
+  /**
+   * How to initialize the struct.
+   */
+  readonly initializer: string;
```
eladb, line 189: Is this a usage example? Maybe each type should have `usage` instead?

src/docgen/schema.ts
```diff
@@ -0,0 +1,335 @@
...
+
+  /**
+   * Instance methods.
+   */
+  readonly instanceMethods: MethodSchema[];
+
+  /**
+   * Static functions.
+   */
+  readonly staticFunctions: MethodSchema[];
```
eladb, line 148:
```suggestion
  readonly staticMethods: MethodSchema[];
```

src/docgen/schema.ts
```diff
@@ -0,0 +1,335 @@
...
+
+  /**
+   * Unique id of the method.
+   */
+  readonly id: string;
+
+  /**
+   * Code snippet to display.
+   */
+  readonly snippet: string;
```
eladb, line 86:
```suggestion
  readonly usage: string;
```

src/docgen/schema.ts
```diff
@@ -0,0 +1,335 @@
...
+  /**
+   * The default value of the property
+   */
+  readonly default?: string;
+}
+
+/**
+ * Describes a parameter.
+ */
+export interface ParameterSchema extends PropertySchema {}
```
eladb, line 71: Parameters are not properties... relation is odd

src/docgen/schema.ts
```diff
@@ -0,0 +1,335 @@
...
+
+  /**
+   * The name of the type (Map, List...)
+   */
+  readonly name: string;
+
+  /**
+   * The various types of the reference.
+   */
+  readonly types?: TypeSchema[];
```
eladb, line 19: Is this for union types?

eladb (changes requested):
Some feedback on the schema. Generally looks good. Would be good to get @rix0rrr 's review on this as well

src/docgen/schema.ts
```diff
@@ -0,0 +1,335 @@
+/**
+ * Describes a type.
+ */
+export interface TypeSchema {
+
+  /**
+   * The type FQN. If missing, the type is a reference (array, map..)
```
rix0rrr, line 7: ...or a primitive?

Is there a reason this type isn't a discriminated union? Feels like it would be much more clear to consume if there was a type field going `type: primitive`, `type: referenceType`, `type: collectionType` etc.

src/docgen/view/class.ts
```diff
@@ -31,19 +32,18 @@ export class Class {
...
-    this.properties = new Properties(transpile, klass.ownProperties, linkFormatter);
+    this.instanceMethods = new InstanceMethods(transpile, klass.ownMethods);
+    this.staticFunctions = new StaticFunctions(transpile, klass.ownMethods);
+    this.constants = new Constants(transpile, klass.ownProperties);
+    this.properties = new Properties(transpile, klass.ownProperties);
     this.transpiled = transpile.class(klass);
   }
 
-  public render(): Markdown {
+  public toMarkdown(linkFormatter: (type: TranspiledType) => string): Markdown {
```
rix0rrr, line 46: For safety's sake, shouldn't we make this a `MarkdownConversionOptions` object? (Which currently has a link formatter, but could have more in the future?

----

Or honestly, it should probably go the other way around:

```ts
class MarkdownRenderer {
  constructor(private readonly linkFormatter: LinkFormatter) {
  }

  public emitClass(klz: ClassSchema) {
    this.emit(`# ${klz.name}`); 
    for (const m of klz.instanceMethods) {
      this.emitInstanceMethod(m);
    }
    
    // ...
  }

}
```

...no?

rix0rrr: (Well I guess that doesn't fit with the current design very well...)

src/docgen/schema.ts
rix0rrr: Thinking about it some more... at this point it looks like the schema is really more of a 1:1 mirror of the jsii assembly. We're losing *some* information but not a lot.

If we're going to retain most of the information from the jsii assembly, then feels like we might as well render from that?

I think the challenge is to lose and compact as much information as we can, while still being able to produce the output that we need. At that point we'll have found the essence of this data model.

I reckon it would maybe looks something a little more like this?

```ts
interface Type {
  id: string;
  name: string;
  jsiiFqn: string;
  summary: string;
  details: string;
  annotations: Annotation[]; // This is where we would stick "deprecated", "experimental", etc
  memberGroups: MemberGroup[]; // Properties, methods, constructors, ...
  // etc
}

interface MemberGroup {
  caption: string;
  members: Member[];
}

interface TypeReference {
  types: SingleType[];
}

interface SingleTypeReference {
  displayName: string;
  link?: Link;
}

type Link =
  | { type: 'reference', id: string }
  | { type: 'href', href: string };


```

rix0rrr: And if we say "no we want it to be mostly the jsii assembly schema but with less data", then feels like the definitions could be a set of `Omit<>`/`Pick<>`-based mapped types...?

src/docgen/view/class.ts
MrArnoldPalmer: yeah we use a lot of positional args on these, but the public API does use an object on [`Documentation.toMarkdown()`](https://github.com/cdklabs/jsii-docgen/pull/481/files#diff-850c9573c26522a8099b803d3478eaf0521a99822c5e1aeec8751de695b9d0baR207)

src/docgen/schema.ts
MrArnoldPalmer: so I do agree with this in principle, and right now it is very close to the assembly, but the main point is that this content is already language specific. So the consumer is avoiding doing all the language specific translations coming from the assembly directly. I see the purpose of the "view model" aka, the next level of translation that we have discussed to be the place where we want to "lose" as much information and present it in a format that can dictate the layout of the content.

src/docgen/schema.ts
MrArnoldPalmer: this is for child types, aka a list of lambda.Code objects in Python appears as
```ts
{
  name: `typing.List[%]`,
  types: [{
    fqn: '@aws-cdk/aws-lambda.Code',
    name: 'aws_lambda.Code',
    types: [],
  }],
}
```
The `%` to represent how to place child types is something I would prefer the consumer to dictate, and was experimenting with allowing the passing of a function similar to `linkFormatter` in the MD. We could change `name` to be `string[]` and pass `['typing.List[', ']']` or something, but basically what clients need is access to the transliterated name for types and fqn, with some indication of how to layout any nested types, but just a giant string isn't good enough cause you need to know where all the discrete types are so you can build appropriate links to named types that may be a subset of the overall signature string that is displayed.

@rix0rrr we discussed this briefly at one point

src/docgen/schema.ts
MrArnoldPalmer: yeah its purely structural, but probably best to just repeat since there is no inherent relationship.

src/docgen/schema.ts
MrArnoldPalmer: Thinking from the frontend app perspective, I was thinking `id` as more generic and explanatory than FQN. But I don't think that's actually the responsibility of this schema so `fqn` I think is more appropriate.

src/docgen/schema.ts
MrArnoldPalmer: `name` is the transliterated name of any given type, vs `id` or `fqn` which is obviously the same across languages.

MrArnoldPalmer:
I'm gonna close this and revisit once more since it's not imminently needed by construct-hub. We should make the JSON model the basis of the markdown generation as discussed with @rix0rrr before. This will change the structure of this quite a bit.

PR #496, opened 2021-11-15
feat: reuse package installation for multiple languages rendering

iliapolo:
Currently the documentation API allows creating a `Documentation` object only for a specific language. This means that if callers are interested in rendering multiple languages, they have to either:

1. Call `Documentation.forPackage` multiple times, resulting in un-necessary package installations.
2. Install the package externally, and use `Documentation.forProject`. This means the package installation code has to be repeated by the caller. This is [exactly](https://github.com/cdklabs/construct-hub/blob/b289fd81ede5d7c5795a76d067ba38ebd8ba1aec/src/backend/transliterator/transliterator.ecstask.ts#L95-L105) what happens in `construct-hub`, which is a shame.

This PR allows reusing package installation for multiple languages by moving the `language` property to the `render` function, and preserving the package installation across calls. 

BREAKING CHANGE: The `language` and `loose` properties have moved from the `Documentation.forXXX` functions to the `Documentation.render` method.


src/docgen/view/documentation.ts
```diff
@@ -145,78 +147,93 @@ export class Documentation {
...
+  public static async forAssembly(assemblyName: string, assembliesDir: string): Promise<Documentation> {
+    return new Documentation(assemblyName, assembliesDir);
   }
 
+  private readonly cleanupDirectories: string[] = [];
+
   private constructor(
-    public readonly assembly: reflect.Assembly,
-    private readonly transpile: Transpile,
+    private readonly assemblyName: string,
```
iliapolo, line 166: We no longer has an `assembly` object since this changes based on the language. I also switched this to `private` since it was exposed for `construct-hub` [usage](https://github.com/cdklabs/construct-hub/blob/main/src/backend/transliterator/transliterator.ecstask.ts#L233), which is not really needed since we know the assembly name in advance.

src/docgen/view/documentation.ts
```diff
@@ -145,78 +147,92 @@ export class Documentation {
...
 
   /**
    * Generate markdown.
    */
-  public render(options?: RenderOptions): Markdown {
-    const submodule = options?.submodule ? this.findSubmodule(this.assembly, options.submodule) : undefined;
+  public async render(options?: RenderOptions): Promise<Markdown> {
+
+    const language = options?.language ?? Language.TYPESCRIPT;
+    const loose = options?.loose ?? true;
```
RomainMuller, line 177: I feel like assigning a default value might make things a little less `?`-heavy... (Feel free to opt out of this).

```suggestion
  public async render(options: RenderOptions = {}): Promise<Markdown> {

    const language = options.language ?? Language.TYPESCRIPT;
    const loose = options.loose ?? true;
```

src/docgen/view/documentation.ts
```diff
@@ -145,78 +147,92 @@ export class Documentation {
...
+    this.cleanupDirectories.push(directory);
+  }
+
+  /**
+   * Removes any internal working directories.
+   */
+  public async cleanup() {
+    for (const dir of this.cleanupDirectories) {
+      await fs.remove(dir);
+    }
```
RomainMuller, line 207: Maybe also remove the entry from `this.cleanupDirectories`? Or drop the list once it's done?

RomainMuller (approved):

src/docgen/view/documentation.ts
iliapolo: I like it

PR #506, opened 2021-11-23
feat: add tables to docs

Chriscbr:
This PR adds tables to the "properties" sections of constructs, classes, and structs so that its easy to see at a glance all of the available properties on a construct and what they do, similar to what is currently available on the AWS CDK docs (see e.g. https://docs.aws.amazon.com/cdk/api/latest/docs/@aws-cdk_aws-s3.Bucket.html).

Examples:
- TypeScript: https://gist.github.com/Chriscbr/198e41bfa19c46574907ee67873abb00
- Python: https://gist.github.com/Chriscbr/1b4415c44b0d26e6d1083abcff9d215e
- Java: https://gist.github.com/Chriscbr/4a7dd9bdbbb6ef0aab3180e4a49c6632
- .NET: https://gist.github.com/Chriscbr/1907a6b44ee2024579366bb164220d59


src/docgen/view/properties.ts
```diff
@@ -17,9 +17,17 @@ export class Properties {
     }
 
     const md = new Markdown({ header: { title: 'Properties' } });
+
+    md.table([
+      ['Name', 'Type', 'Description'].map(Markdown.bold),
+      ...this.properties.map((prop) => [prop.name, prop.type, Markdown.sanitize(prop.description)]),
```
eladb, line 23: Can we make the "name" link to the property section?

eladb (changes requested):

Chriscbr (commented):
Also just realized I can add tables to other sections as well, this is done in the AWS CDK API reference

src/docgen/render/markdown.ts
```diff
@@ -177,7 +193,7 @@ export class Markdown {
           `${heading} <span data-heading-title="${this.header}" data-heading-id="${anchor}"></span>`,
         );
       } else {
-        content.push(`${heading} <a name="${this.id}"></a>`);
+        content.push(`${heading} <a name="${this.id}" id="${anchor}"></a>`);
```
eladb, line 196: I am not sure but I think that maybe we want the anchor just before the heading so when a link is clicked the viewport will scroll so the heading is visible. I never figured out the right way to do this. Maybe @gabewomble has some insight

eladb (approved):
Looks great. Approved and added do-not-merge

src/docgen/render/markdown.ts
Chriscbr: It seems like the headings are visible when I click on anchored links on the gist or on Construct Hub, is this maybe a browser specific thing? Maybe create an issue about it.

If I understand your issue correctly, it might be better fixed with some CSS perhaps https://stackoverflow.com/questions/11501025/div-anchors-scrolling-too-far

Chriscbr:
Updated the PR description with a list of examples in each language.

PR #533, opened 2021-12-24
feat!: JSON doc generation

Chriscbr:
Supersedes #481
Part of https://github.com/cdklabs/construct-hub-webapp/issues/714 and https://github.com/cdklabs/construct-hub/issues/142

Allows rendering language specific documentation to a JSON object in addition to a markdown string. This allows the consumer more flexibility to render the structure of the API reference as needed without regenerating markdown documents.

This PR builds upon #481 by cleaning up the JSON schema (in `schema.ts`) and also changing the markdown rendering logic to be based on the JSON output -- so the markdown rendering does not depend on any `TranspiledXxx` classes or interfaces.

All headers now have anchor tags which default to the unique language-agnostic FQNs (e.g. `@aws-cdk/aws-s3.Bucket.parameter.scope`). Based on my research, all characters used in JSII FQNs (A-Za-z0-9_-/.@) are allowed as URL fragments, and are also allowed as id's in HTML5, so this is a reasonable default.

This PR also extends the notion of fully qualified names to all "entities" in the document (methods, properties, enum members, and even method parameters) denoted by the "id" fields within the JSON schema, so that any section of a document can be uniquely linked - not just types.  The `JsiiEntity` interface in `schema.ts` is intended to encapsulate any information that should be needed to render a link, including the package version it's from (this is necessary for Construct Hub to be able to link to the correct versions of packages).

Markdown generation can now be customized using three hooks:

- `anchorFormatter: (type: JsiiEntity) => string` - customize the IDs given to anchors tags that are placed next to every header in the document, so that any part of the API reference can be directly linked
- `linkFormatter: (type: JsiiEntity, metadata: AssemblyMetadataSchema) => string` - customize how links should be rendered when they appear in tables, descriptions, etc. -- e.g. to include logic to "link to another page"
- `typeFormatter?: (type: TypeSchema, linkFormatter) => string` - customize how composite types should be rendered when they include nested types - this allows functionality like the screenshot below:

Screen Shot 2021-12-23 at 7 34 14 PM

See `markdown.test.ts` for an example of how the markdown rendering hooks would be used to format links for Construct Hub.

Examples:

@aws-cdk/aws-ecr Python API.json: https://gist.github.com/Chriscbr/731b315808faaf2a2161d686ce248368
@aws-cdk/aws-ecr Python API.md: https://gist.github.com/Chriscbr/d0833a9ecce258dead648a63253b8b0e

BREAKING CHANGE: `Documentation.render` is now named `Documentation.toMarkdown`. In addition, the options parameter for this method has changed. `TranspiledType` is no longer exported.


Chriscbr:
Gist with sample docs from Python: https://gist.github.com/Chriscbr/1e71b60bd76394da3e285e05a62c1aee

One opinionated change I made was to change how code is rendered, from the usual markdown syntax to using HTML < pre > (multi-line) and < code > (single-line) tags. Doing this allows us to include links within snippets even when they're rendered on GitHub, like so:

Screen Shot 2021-12-23 at 7 34 14 PM

src/docgen/render/markdown.ts
```diff
@@ -180,21 +181,8 @@ export class Markdown {
...
-      // Note though that cross links (i.e links dependencies will not work yet regardless)
-      const headerSpan = !!process.env.HEADER_SPAN;
-      if (headerSpan) {
-        content.push(
-          `${heading} <span data-heading-title="${this.header}" data-heading-id="${anchor}"></span>`,
-        );
-      } else {
-        content.push(`${heading} <a name="${this.id}" id="${anchor}"></a>`);
-      }
+      content.push(`${heading} <a id="${this.id}"></a>`);
```
Chriscbr, line 185: Currently the id's that are generated in the schema are the full JSII FQNs - e.g. `@aws-cdk/aws-s3.BucketProps.parameter.accessControl`. But as mentioned in https://github.com/cdklabs/construct-hub-webapp/issues/714, it would be nice to just use assembly-relative IDs for URL fragments in Construct Hub (i.e. just `BucketProps.parameter.accessControl`).

For example, our current URLs look like:
`https://constructs.dev/packages/@aws-cdk/aws-s3/v/1.137.0/api/softwareamazonawscdkservicess3bucket#softwareamazonawscdkservicess3bucketpropsparameterbucketname`
and with the current state of the PR, we can probably make it
`https://constructs.dev/packages/@aws-cdk/aws-s3/v/1.137.0/api/Bucket#@aws-cdk/aws-s3.BucketProps.parameter.bucketName`
but we could optimize it further and make it
`https://constructs.dev/packages/@aws-cdk/aws-s3/v/1.137.0/api/Bucket#BucketProps.parameter.bucketName`

That is probably what we want at the end of the day, so I need to update the IDs used by markdown rendering to be customizable. TODO

rix0rrr:
> One opinionated change I made was to change how code is rendered by default, from the usual markdown syntax to using HTML pre (multi-line) and code (single-line) tags. Doing this allows us to include links within snippets even when they're rendered on GitHub, as shown above.

I want to say: what's the point of MarkDown anymore at this stage 😅 but I support the usability improvements

src/docgen/render/markdown.ts
```diff
@@ -233,3 +220,46 @@ export class Markdown {
...
   }
 }
+
+export const defaultAnchorFormatter = (type: JsiiEntity) => {
+  return type.id;
+};
+
+export const defaultLinkFormatter = (type: JsiiEntity) => {
+  const name = type.fqn.split('.').pop()!;
+  return `<a href="#${type.id}">${name}</a>`;
```
rix0rrr, line 230: What if the type comes from a different assembly?

src/docgen/render/markdown.ts
```diff
@@ -233,3 +220,46 @@ export class Markdown {
...
+};
+
+function isJsiiType(value: any): value is JsiiEntity {
+  return typeof value === 'object' && value.fqn && value.id;
```
rix0rrr, line 234: This may explode if `value` is ever `null`. Thanks, JavaScript!

src/docgen/schema.ts
```diff
@@ -0,0 +1,428 @@
...
+  /**
+   * The language-specific name of the type. May contain "%" placeholder
+   * values to indicate references to types defined in the "types" field.
+   *
+   * @example "string"
+   * @example "%"
+   * @example "typing.List[%]"
+   * @example "Map<%, %>"
+   */
+  readonly name: string;
```
rix0rrr, line 15: `name` doesn't feel like a great name if this is more like a `formattingPattern`, especially since for simple jsii types it seems the name is never in here.

src/docgen/schema.ts
```diff
@@ -0,0 +1,428 @@
...
+
+/**
+ * Describes a single "entity" in the jsii type system. This may be a type,
+ * but it could also be a property, method, parameter, enum member, etc.
+ */
+export interface JsiiEntity extends AssemblyMetadataSchema {
+  /**
+   * The language-specific type FQN.
+   */
+  readonly fqn: string;
```
rix0rrr, line 31: How will we get to a simple `displayName` here? Is that always the last `.`-separated part of the `fqn` ?

Because that might be incorrect in languages in which the `fqn` is not `.`-separated, or in which we want to have more than one component part of the displayName.

Better to make it a separate field I think.

src/docgen/schema.ts
```diff
@@ -0,0 +1,428 @@
...
+export interface CallableSchema extends Usage {
+  /**
+   * An id that uniquely identifies this callable among all entities in
+   * the document and is same across languages.
+   */
+  readonly id: string;
+
+  /**
+   * The language-specific fqn. The last part after splitting by "." is
+   * the method name.
```
rix0rrr, line 103: Also for initializers?

src/docgen/view/class.ts
```diff
@@ -10,6 +13,48 @@ import { StaticFunctions } from './static-functions';
 const CONSTRUCT_CLASS = 'constructs.Construct';
 
 export class Class {
+  public static toMarkdown(
```
rix0rrr, line 16: It looks like this class now holds both 

* Conversion to JSON (member function)
* MarkDown rendering (static function)

This seems like 2 distinct concerns, and feels like they shouldn't be combined. The fact that the MarkDown conversion methods are stateless also requires passing around the `MarkdownRenderContext` object everywhere.

Why don't we make the Markdown rendering a (visitor) class instead:

```ts
public class MarkdownRenderer {
  private readonly document = new MarkdownDocument();

  constructor(private readonly linkFormatter, private readonly otherFormatter, private readonly someOtherContext) {
  }

  public visitClass(klass: ClassSchema) {
    this.document.h2(klass.name);
    this.document.paragraph(klass.docs.summary);
    this.document.paragraph(klass.docs.remarks);

    for (const method of klass.methods) {
      this.visitMethod(method);
    }
  }

  public visitMethod(method: MethodSchema) { }
}
```

This keeps all the MD concerns nicely colocated, and all renderers can just happily share state and private helpers.

rix0rrr (changes requested):
I like what you're doing, and I appreciate the effort. I'd prefer to see the MarkDown renderer factored out into its own visitor class. Seems like a neater separation of concerns, makes it easier to add new types of renders, and more in line with the final processing pipeline.

src/docgen/schema.ts
Chriscbr: Hmm, yeah I agree that could be a risky assumption to make. I'll add in `displayName` 👍

src/docgen/schema.ts
Chriscbr: Updated

src/docgen/render/markdown.ts
Chriscbr: In that case we just shrug it off. This is just a default after all. The id is unique across assemblies, so in the worst case the link is invalid. You can see this in the example gist [here](https://gist.github.com/Chriscbr/d0833a9ecce258dead648a63253b8b0e), where if you click on a type named `Construct`, the link doesn't take you anywhere.

I think this is good enough for the default.

src/docgen/render/markdown.ts
Chriscbr: `typeof null === 'object'`... TIL!

src/docgen/schema.ts
Chriscbr: Fixed

src/docgen/render/markdown.ts
Chriscbr: Actually, I suppose we could display no link if the type isn't from the current assembly. I'll leave this as a future enhancement for now since it's not too critical

Chriscbr:
@rix0rrr I've given a shot at refactoring all of the markdown-ification into a visitor class - definitely feels like an improvement. Let me know what you think.

src/docgen/render/markdown.ts
rix0rrr: See I disagree with this.

A link that doesn't work seems worse than no link at all. I'd rather have that be in the initial implementation.

rix0rrr (changes requested):
I think I'd like to see the links changed. Let's not generate links that don't go anywhere.

Other than that, <3 this and ready to ship it!

README.md
```diff
@@ -17,11 +17,18 @@ As a library:
 ```ts
 import { Documentation } from 'jsii-docgen';
 
-const docs = await Documentation.forLocalPackage('.', { language: 'ts' });
-const markdown = docs.render().render(); // returns a markdown string
+const docs = await Documentation.forProject('.');
+const markdown = docs.toMarkdown({ language: 'typescript' }).render(); // returns a markdown string
```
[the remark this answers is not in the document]
iliapolo: I believe this should be:

```suggestion
const markdown = await docs.toMarkdown({ language: Language.TYPESCRIPT }).render(); // returns a markdown string
```

src/docgen/render/markdown.ts
Chriscbr: Got it, I see your point now - yeah dead links are a pretty bad experience. 😅

I've added another parameter to the `linkFormatter` API to support this -- as of right now, `linkFormatter` is expected to be a pure function, so the formatter needs extra info to make the determination of whether to create a link.

The formatting APIs are getting a bit heavy/awkward so I've added some more docs and marked them experimental for now. It might make sense eventually just require anyone that wants to customize the markdown linking and formatting etc. to just implement their own markdown renderer at the end of the day -- but for now I'd rather keep these (awkward) APIs just to avoid duplicating code between here and Construct Hub.

Chriscbr:
I'm getting timed out trying to create or update gists on GitHub - but here's a screenshot of the "fixed" links - for types that are not in the current document, we just show the full language-specific fqn.

Screen Shot 2022-01-12 at 12 45 18 PM

test/docgen/view/__snapshots__/class.test.ts.snap
```diff
@@ -1,32 +1,87 @@
...
+      },
+      "fqn": "Amazon.CDK.AWS.ECR.AuthorizationToken.GrantRead",
+      "id": "@aws-cdk/aws-ecr.AuthorizationToken.grantRead",
+      "parameters": Array [
+        Object {
+          "default": undefined,
+          "displayName": "Grantee",
+          "docs": Object {},
+          "fqn": "Amazon.CDK.AWS.ECR.AuthorizationToken.parameter.Grantee",
+          "id": "@aws-cdk/aws-ecr.AuthorizationToken.parameter.grantee",
```
iliapolo, line 31: I'm not sure of the implications here, but shouldn't this be:

```suggestion
          "id": "@aws-cdk/aws-ecr.AuthorizationToken.grandRead.parameter.grantee",
```

If this is being used to render links, if two methods have the same parameter, this might get confusing.

src/docgen/schema.ts
```diff
@@ -0,0 +1,465 @@
...
+  /**
+   * Enums.
+   */
+  readonly enums: EnumSchema[];
+}
+
+/**
+ * Metadata about a particular jsii assembly.
+ */
+export interface AssemblyMetadataSchema {
```
iliapolo, line 344: I feel like we should add a `language` property here. Consumers should be able to tell which language they are looking at.

iliapolo (changes requested):
Looks great!

test/docgen/view/__snapshots__/class.test.ts.snap
```diff
@@ -1,32 +1,87 @@
...
 
 Authorization token to access private ECR repositories in the current environment via Docker CLI.
 
 > https://docs.aws.amazon.com/AmazonECR/latest/userguide/registry_auth.html
 
+> [https://docs.aws.amazon.com/AmazonECR/latest/userguide/registry_auth.html](https://docs.aws.amazon.com/AmazonECR/latest/userguide/registry_auth.html)
 
-# Static Functions <a name=\\"Static Functions\\" id=\\"static-functions\\"></a>
+
+# Static Functions <a name=\\"Static Functions\\" id=\\"Static Functions\\"></a>
```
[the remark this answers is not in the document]
Chriscbr: Ow, yeah that's no good. Fixed.

test/docgen/view/__snapshots__/documentation.test.ts.snap
```diff
@@ -55687,7 +55687,7 @@ aws_cdk.aws_ecr.CfnRepository(
                     "fqn": "constructs.Construct",
                     "id": "constructs.Construct",
                     "packageName": "constructs",
-                    "packageVersion": "3.3.187",
+                    "packageVersion": "3.3.188",
```
Chriscbr, line 55690: Hrmmmmm. It looks like what's happening is that since most packages we generate have caret dependencies, the version of the referenced type isn't necessarily fixed.

I've been thinking about this for a while, including whether there's some better way to model this so that the package version stays fixed given the current assembly (e.g. by including the semver string from package.json instead) -- but it turns out that forces a lot of additional complexity onto downstream users.

For now I'm going to simply add some extra code that will make the snapshots stable by always linking to the same package version within unit tests (e.g. "1.2.3")

src/docgen/transpile/transpile.ts
```diff
@@ -425,6 +425,15 @@ export class TranspiledTypeReference {
       if (!this.ref.fqn) {
         throw new Error(`Original type reference for ${this.type.fqn} does not have a fqn.`);
       }
+
+      // If we are running in a test, report a fake stable version since types
+      // may belong to dependency packages, which could be specified with caret
+      // dependencies - this means the packageVersion of a type like
+      // `construct.Construct` will be set to whichever version is installed.
+      // This is okay in practice, but makes snapshot tests inconsistent.
```
iliapolo, line 433: Why is this a problem though? Our snapshots will get updated during self-mutation

iliapolo: Scratch that, its inconsistent over time you mean, unrelated to dependency upgrades 👍

rix0rrr (approved):

iliapolo (approved):

PR #553, opened 2022-01-11
chore: cache expensive lookups

iliapolo:
Currently, ingesting packages with a large amount of submodules and/or types can take a very long time.

For example, running the following:

```ts
const docs = await Documentation.forPackage('@cdktf/provider-aws@4.0.1');
console.time('render');
await docs.render({ language: Language.PYTHON, submodule: 'wafv2' });
console.timeEnd('render');
```

Will result in:

```console
Installing package @cdktf/provider-aws@4.0.1
render: 9:52.761 (m:ss.mmm)
```

With this PR, by adding a couple of caches, this is reduced to:

```console
Installing package @cdktf/provider-aws@4.0.1
render: 32.382s
```

Fixes https://github.com/cdklabs/construct-hub/issues/664


src/docgen/transpile/transpile.ts
```diff
@@ -667,11 +670,16 @@ export abstract class TranspileBase implements Transpile {
   }
 
   protected getParentModule(moduleLike: reflect.ModuleLike): reflect.Assembly {
-    if (moduleLike.types.length === 0) {
+    const cached = this.parentModulesCache.get(moduleLike.fqn);
+    if (cached) return cached;
+
+    const types = moduleLike.types;
```
iliapolo, line 676: This was the culprit. Apparently doing [`Object.values`](https://github.com/aws/jsii/blob/main/packages/jsii-reflect/lib/module-like.ts#L37) on a map with 8000 entries takes roughly 2.3ms, and this used to happen twice for each `typeReference` transpilation. 

Specifically for non typescript languages this is exacerbated because of parameter expansion.

iliapolo:
Some more analysis details. Before this PR, here is how the flamegraph looked like:

Screen Shot 2022-01-11 at 9 58 51 PM

We can see the `render` method is taking the bulk amount of time compared to `loadAssembly`. 
It also shows very noticeable hotspots 

After this PR:

Screen Shot 2022-01-11 at 9 59 09 PM

We can see `render` is now a small fraction of the cost, with no substantial hotspots on the call stack. 
The hottest spot now is the `validateSchema` function of `jsonschema`. 

@RomainMuller Do you think we can improve something there, or maybe run `loadAssembly` without validation? I don't think we need validation at this point, at least not where its used in construct hub.

src/docgen/transpile/transpile.ts
```diff
@@ -642,36 +646,48 @@ export abstract class TranspileBase implements Transpile {
...
-    const submoduleFqn = `${type.assembly.name}.${type.namespace.split('.')[0]}`;
-    const submodules = type.assembly.submodules.filter(
-      (s) => s.fqn === submoduleFqn,
-    );
+    let submodule = undefined;
+    if (type.namespace) {
+      // if the type is in a submodule, the submodule name is the first
+      // part of the namespace. we construct the full submodule fqn and search for it.
+      const submoduleFqn = `${type.assembly.name}.${type.namespace.split('.')[0]}`;
+      const submodules = type.assembly.submodules.filter(
```
iliapolo, line 659: According to the flame graph, the `assembly.submodules` call is also pretty expensive. The submodules cache here reduced another 15 seconds from the render time.

MrArnoldPalmer (approved):
Awesome! 🔥 🔥 🔥

PR #559, opened 2022-01-14
chore: disable assembly validation

Chriscbr:
To improve performance, we change the default behavior to disable validating assemblies while loading them with `jsii-reflect`. This matches the behavior that already exists to skip assembly validations when we invoke `transliterateAssembly` from `jsii-rosetta`, as seen in this line here: https://github.com/aws/jsii/blob/309da85ba07c8ac884dcfb72fb0cbc0271364326/packages/jsii-rosetta/lib/commands/extract.ts#L89

Validation can be re-enabled via the `validate` flag.

When generating python docs for the root module of `aws-cdk-lib@2.8.0`, the time it took to generate on my development machine decreased from 30 seconds to 16 seconds.

Before flamegraph:
Screen Shot 2022-01-14 at 2 39 06 PM

After flamegraph:
Screen Shot 2022-01-14 at 2 39 40 PM

### DIFF
--- a/src/docgen/view/documentation.ts
+++ b/src/docgen/view/documentation.ts
@@ -25,19 +25,11 @@ const NOT_FOUND_IN_ASSEMBLY_REGEX = /Type '(.*)\..*' not found in assembly (.*)$
 /**
  * Options for rendering a `Documentation` object.
  */
-export interface RenderOptions {
+export interface RenderOptions extends TransliterationOptions {
   /**
    * Which language to generate docs for.
    */
   readonly language: Language;
-
-  /**
-   * Whether to ignore missing fixture files that will prevent transliterating
-   * some code snippet examples.
-   *
-   * @default true
-   */
-  readonly loose?: boolean;
 
   /**
    * Include a generated api reference in the documentation.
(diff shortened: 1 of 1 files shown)


eladb (commented):
No test?

src/docgen/view/documentation.ts
```diff
@@ -245,9 +256,9 @@ export class Documentation {
     }
   }
 
-  private async languageSpecific(lang: Language, loose: boolean): Promise<{ assembly: reflect.Assembly; transpile: Transpile}> {
+  private async languageSpecific(lang: Language, options: TransliterationOptions): Promise<{ assembly: reflect.Assembly; transpile: Transpile}> {
```
iliapolo, line 259: I actually don't like reusing `TransliterationOptions` here because it might encourage duplicate handling of default values. 

This method, while it accepts the same arguments, actually uses a different type, because it intentionally forces those arguments to be required.

i.e

```ts
export interface A {
  readonly y: string;
}
```

is a different type than

```ts
export interface A {
  readonly y?: string;
}
```

Chriscbr: Since the APIs are private, I updated the parameters to use `Required<>` where appropriate to make this more clear

Chriscbr:
> No test?

I'm not sure how I would test this really. There's no difference in "outputs" that I see that would be easy to measure. (open to ideas)

I think the proof is in the pudding - the total time to run all tests decreased from 10 minutes to about 2.5 minutes on my machine (with comparable improvements on the GitHub runner) 🔥

MrArnoldPalmer (approved):

PR #571, opened 2022-01-22
fix: invalid values for submodule in JSON output

### LINKED ISSUE #570: Invalid values for submodule in JSON output
I ran `jsii-docgen@5.0.5` on `@aws-solutions-constructs/aws-lambda-sqs@2.0.0` with Java selected as the target language, and found this fragment:

```
{
  "fqn": "software.amazon.awsconstructs.services.lambdasqs.LambdaToSqs.Initializer.parameter.deadLetterQueueProps",
  "displayName": "deadLetterQueueProps",
  "id": "@aws-solutions-constructs/aws-lambda-sqs.LambdaToSqs.Initializer.parameter.deadLetterQueueProps",
  "optional": true,
  "default": "- Default props are used",
  "type": {
    "formattingPattern": "%",
    "types": [
      {
        "id": "aws-cdk-lib.aws_sqs.QueueProps",
        "displayName": "QueueProps",
        "fqn": "software.amazon.awscdk.services.sqs.QueueProps",
        "packageName": "aws-cdk-lib",
        "packageVersion": "2.8.0",
        "submodule": "software.amazon.awscdk.services.sqs"
      }
    ]
  },
  "docs": {
    "summary": "Optional user provided properties for the dead letter queue."
  }
},
```

Based on how the schema was designed, `submodule` should be equal to `aws_sqs` (the language-independent module name), not `software.amazon.awscdk.services.sqs`.

Chriscbr:
Fixes #570

The root cause of the bug was that when calling `toJson` in `transpile.ts`, we were generating the value of `submodule` from `TranspiledType`, which actually refers to the _transliterated_ submodule name. We fix this by creating a new field called `submoduleFqn` that stores the correct submodule name, and using that instead.

I've added extra unit tests with additional package + language combinations that previously failed to prevent regressions.

This PR also fixes a bug where transliterating a corrupted assembly did not properly throw `CorruptedAssemblyError`. The root cause of this was that we were doing some transpiling within the `toJson()` methods instead of within class constructors, and our try-catch block only covered the initialization of the `ApiReference` class. I fixed this by refactoring so all transpiling happens within constructors. (This was already the case for most classes, there were just a few exceptions where this wasn't being followed).


src/docgen/schema.ts
```diff
@@ -493,7 +497,16 @@ export function extractDocs(docs: jsii.Docs): DocsSchema {
   });
 }
 
-function filterUndefined<T>(obj: T): T {
+/**
+ * Generates the name of the submodule.
```
iliapolo, line 501: Is this really extracting the name? The function seems to suggest it extracts path, and its usage suggest it extract an FQN.

Can you align all these? I think `submodulePath` actually correctly represents what this function is doing, but in that case I would change the schema to declare `submodulePath` as well, not `submoduleFqn`.

iliapolo (approved):
Just one comment, feel free to remove the `do-not-merge` label once you address it.

PR #577, opened 2022-01-26
fix!: CLI only generates root module by default

### LINKED ISSUE #540: CLI command only generates docs in root submodule by default
JSII projects may have multiple submodules (for examples libraries, see `aws-cdk-lib` and `projen`). Currently the CLI only generates docs for the root submodule by default. This appears to be a regression from jsii docgen v0.1.x.

I think the default behavior for both the programmatic API and the CLI should be to render docs for all submodules if no other option has been specified. There should also be a way in the CLI to specify specific submodules to render (to match the `submodule` option available in the programmatic API).

Chriscbr:
Fixes #540

No changes have been made to the defaults of the programmatic API. To get this same behavior (where all APIs are rendered), there is a new boolean option called `allSubmodules`.

BREAKING CHANGE: The jsii-docgen CLI now generates documents with APIs from all jsii submodules by default. To only generate API.md for a specific submodule, pass the `--submodule <submod>` option, or pass `--submodule root` to only render APIs from the root module.


src/docgen/render/json.ts
```diff
@@ -4,7 +4,7 @@
 export class Json<T> {
   constructor(public readonly content: T) {};
 
-  public render(replacer?: any, space?: string | number): string {
-    return JSON.stringify(this.content, replacer, space);
+  public render(): string {
+    return JSON.stringify(this.content, null, 2);
```
Chriscbr, line 8: I included these options in the original API, but in retrospect it's a bit clunky and I think it makes more sense to render the document with indentation by default. If someone wants the old behavior they can just access the public `content` field directly

MrArnoldPalmer:
It seems like we should bump the major version because of the breakage to the CLI behavior?

Chriscbr:
Hrm... yeah I think that's a good call. Alright I've updated the PR description 👍

MrArnoldPalmer (approved):

PR #588, opened 2022-02-02
fix: anchor tag for Initializer isn't customized

Chriscbr:
All of the anchor tags generated within the `MarkdownRenderer` class should be updated if a custom `anchorFormatter` is specified, but we forgot to call `anchorFormatter` when generating the struct initializer header.

### DIFF
--- a/src/docgen/render/markdown-render.ts
+++ b/src/docgen/render/markdown-render.ts
@@ -229,7 +229,12 @@ export class MarkdownRenderer {
     }
 
     const initializer = new MarkdownDocument({
-      id: `${struct.id}.Initializer`,
+      id: this.anchorFormatter({
+        id: `${struct.id}.Initializer`,
+        displayName: 'Initializer',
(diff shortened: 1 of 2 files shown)


MrArnoldPalmer:
@Chriscbr does this break when linking to types in a different package?

Chriscbr:
@MrArnoldPalmer good question, it shouldn't. By default the anchor formatter will just return the entire id back, but the customization hooks can let you choose what happens if the linked type is in a different package.

So in this PR a snapshot got updated for a test case where we give a custom formatter (https://github.com/cdklabs/jsii-docgen/blob/c5fa457ca332863a2384522f46fca3bf6fb99979/test/docgen/view/markdown.test.ts), but none of the other snapshots updated, which seems appropriate.

This change is pretty minor since we don't currently link to initializers much anyways, but still seemed worth fixing

MrArnoldPalmer (approved):

PR #589, opened 2022-02-02
fix: properties and methods from superclasses are not rendered

### LINKED ISSUE #586: Methods and properties from superclasses are not generated in docs
In the AWS CDK, the class `aws_s3.Bucket` extends from `aws_s3.BucketBase`, and correspondingly inherits several methods and properties. These methods and properties are currently not included in the generated docs, even though they should be.

Chriscbr:
Fixes #586

### DIFF
--- a/src/docgen/view/class.ts
+++ b/src/docgen/view/class.ts
@@ -36,10 +36,10 @@ export class Class {
     if (klass.initializer) {
       this.initializer = new Initializer(transpile, klass.initializer);
     }
-    this.instanceMethods = new InstanceMethods(transpile, klass.ownMethods);
-    this.staticFunctions = new StaticFunctions(transpile, klass.ownMethods);
-    this.constants = new Constants(transpile, klass.ownProperties);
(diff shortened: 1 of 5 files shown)


Chriscbr:
I think this is the better experience for customers, since it's not currently obvious that e.g. `Bucket` extends from `BucketBase` and thus has access to all of the same methods. And even if we did explicitly linked to superclasses within generate docs, IMO listing all of available methods / properties for each class is still a better UX.

This does significantly increase the size of docs (noting from the increased size of snapshots in this PR), but I don't think the time to generate docs increased comparatively as much, so I don't expect it to cause downstream issues on Construct Hub for example. But we'll see.

RomainMuller (approved):
It'd be nice to also mark those methods as inherited in the markdown...

Chriscbr:
> It'd be nice to also mark those methods as inherited in the markdown...

Good suggestion! I create an issue as to leave that as an enhancement for now :)

PR #617, opened 2022-02-26
feat: render transliterated code examples

### LINKED ISSUE #585: Include jsii-rosetta examples
The AWS CDK API reference includes example code snippets for most constructs, classes, structs, and enums. [Example](https://docs.aws.amazon.com/cdk/api/v2/docs/aws-cdk-lib.aws_s3.Bucket.html#example). These are generated from information within jsii "tablet" files (if you install `aws-cdk-lib`, within the package you should find a `.jsii.tabl.json` file).

However, these are not displayed when you generate an API reference with jsii-docgen (and hence, they're not shown on Construct Hub). [Example](https://constructs.dev/packages/aws-cdk-lib/v/2.10.0/api/Bucket?lang=typescript&submodule=aws_s3). We should find a way to include these since many users find them helpful given that CDK is vended in multiple languages.

Chriscbr:
Fixes #585

Example gist: https://gist.github.com/Chriscbr/7b4bf9180b7fea165f741a42c5307edd

### DIFF
--- a/src/docgen/render/markdown-doc.ts
+++ b/src/docgen/render/markdown-doc.ts
@@ -102,7 +102,7 @@ export class MarkdownDocument {
   /**
    * Render a docs element into the markdown.
    */
-  public docs(docs: DocsSchema, language: Language) {
+  public docs(docs: DocsSchema, language?: Language) {
     if (docs.summary) {
       this.lines(MarkdownDocument.sanitize(docs.summary));
       this.lines('');
@@ -119,6 +119,9 @@ export class MarkdownDocument {
     }
 
     if (docs.example) {
+      if (!language) {
+        throw new Error('language must be provided if docs.example has been specified');
+      }
       const example = new MarkdownDocument({
         id: `${this.options.id}.example`,
         header: {
--- a/src/docgen/render/markdown-doc.ts
+++ b/src/docgen/render/markdown-doc.ts
@@ -1,3 +1,4 @@
+import { Language } from '../..';
 import { DocsSchema } from '../schema';
 
 /**
@@ -101,7 +102,7 @@ export class MarkdownDocument {
   /**
    * Render a docs element into the markdown.
    */
-  public docs(docs: DocsSchema) {
+  public docs(docs: DocsSchema, language: Language) {
     if (docs.summary) {
       this.lines(MarkdownDocument.sanitize(docs.summary));
       this.lines('');
@@ -115,6 +116,18 @@ export class MarkdownDocument {
       for (const link of docs.links) {
         this.quote(`[${link}](${link})`);
       }
+    }
+
+    if (docs.example) {
+      const example = new MarkdownDocument({
+        id: `${this.options.id}.example`,
+        header: {
+          title: 'Example',
+        },
+      });
+      example.code(language.toString(), docs.example);
+      example.lines('');
+      this.section(example);
     }
   }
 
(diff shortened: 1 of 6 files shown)


test/docgen/view/__snapshots__/enum.test.ts.snap
```diff
@@ -32,23 +32,25 @@ Object {
 exports[`csharp snapshot 2`] = `
 " TagMutability <a name=\\"TagMutability\\" id=\\"@aws-cdk/aws-ecr.TagMutability\\"></a>
 
+The tag mutability setting for your repository.
+
+# Enum Members <a name=\\"Enum Members\\" id=\\"Enum Members\\"></a>
```
Chriscbr, line 37: Because of the way a `MarkdownDocument`'s "sections" are always rendered after their "lines", I noticed example snippets would get rendered after the table listing all of the available enums. The cleanest fix I saw for this was to put the enum members + table in its own section, similar to what we do for classes/structs/interfaces/etc. - though it did cause some extra changes in the snapshots.

test/docgen/view/__snapshots__/documentation.test.ts.snap
```diff
@@ -13807,6 +13823,14 @@ This is the URL inside the kubeconfig file to use with kubectl
 
 ---
 
+###### Example <a name=\\"Example\\" id=\\"aws-cdk-lib.aws_eks.Cluster.property.clusterEndpoint.example\\"></a>
+
+\`\`\`csharp
+// Example automatically generated from non-compiling source. May contain errors.
+https:
+\`\`\`
```
Chriscbr, line 13831: This looks incorrect, but I did some digging and found that the reason this example (and a few others) are being rendered as "code examples" is just due to errors in the older version of `aws-cdk-lib` that's being used as a test fixture, and the examples appear correctly in newer versions of aws-cdk-lib.

It's probably worth updating these test fixtures to newer ones, but I'll leave that open for a later PR.

src/docgen/render/markdown-doc.ts
```diff
@@ -101,7 +102,7 @@ export class MarkdownDocument {
   /**
    * Render a docs element into the markdown.
    */
-  public docs(docs: DocsSchema) {
+  public docs(docs: DocsSchema, language: Language) {
```
iliapolo, line 105: This is technically a breaking change...I think it should be optional because the `example` in `DocsSchema` is also optional, and the language is only needed for the example.

```suggestion
  public docs(docs: DocsSchema, language?: Language) {
```

src/docgen/render/markdown-render.ts
```diff
@@ -337,17 +337,30 @@ export class MarkdownRenderer {
...
+  }
+
+  public visitEnumMembers(
+    enus: EnumMemberSchema[],
+  ): MarkdownDocument {
+    if (enus.length === 0) {
+      return MarkdownDocument.EMPTY;
     }
 
+    const md = new MarkdownDocument({ header: { title: 'Enum Members' } });
```
iliapolo, line 356: "Enum" is redundant at this point.

```suggestion
    const md = new MarkdownDocument({ header: { title: 'Members' } });
```

iliapolo (changes requested):

src/docgen/render/markdown-doc.ts
Chriscbr: Ooh, good catch! 👍

iliapolo (approved):

PR #641, opened 2022-03-21
fix: header limit reached on some packages with examples

Chriscbr:
On some packages, we are seeing an error where jsii-docgen throws an error

```Error: Unable to render markdown. Header limit (6) reached.
    at MarkdownDocument.render (/Volumes/workplace/jsii-docgen/lib/docgen/render/markdown-doc.js:105:19)
```

This is caused by the existence of examples on parameters of methods in some libraries. To see concretely the chain of header sizes, I found that for `@aws-cdk/aws-apigateway@1.47.0` it was trying to render:

```
# API Reference <a name="API Reference" id="api-reference"></a>

## Constructs <a name="Constructs" id="Constructs"></a>

### ProxyResource <a name="ProxyResource" id="@aws-cdk/aws-apigateway.ProxyResource"></a>

#### Methods <a name="Methods" id="Methods"></a>

##### `add_method` <a name="add_method" id="@aws-cdk/aws-apigateway.ProxyResource.addMethod"></a>

###### `request_models`<sup>Optional</sup> <a name="request_models" id="@aws-cdk/aws-apigateway.ProxyResource.addMethod.parameter.requestModels"></a>

####### Example <a name="Example" id="@aws-cdk/aws-apigateway.ProxyResource.addMethod.parameter.requestModels.example"></a>
```

Since H7 is not allowed, we change examples to be rendered without a header, and instead with the title "Example" italicized.

### DIFF
--- a/src/docgen/render/markdown-doc.ts
+++ b/src/docgen/render/markdown-doc.ts
@@ -124,10 +124,8 @@ export class MarkdownDocument {
       }
       const example = new MarkdownDocument({
         id: `${this.options.id}.example`,
-        header: {
-          title: 'Example',
-        },
       });
+      example.lines(MarkdownDocument.italic('Example'), '');
       example.code(language.toString(), docs.example);
       example.lines('');
       this.section(example);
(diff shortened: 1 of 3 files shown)


src/docgen/render/markdown-doc.ts
```diff
@@ -174,15 +172,16 @@ export class MarkdownDocument {
...
-
     const content: string[] = [];
+
     if (this.header) {
+      if (headerSize > 6) {
+        // headers are mapped to `h1-h6` html elements.
+        // passed that, markdown just renders `#` signs.
+        // lets see if and when we'll hit this limit.
+        throw new Error('Unable to render markdown. Header limit (6) reached.');
+      }
```
Chriscbr, line 183: It only makes sense to throw this error if we're actually rendering a header

otaviomacedo (approved):

PR #643, opened 2022-03-21
fix: remove newlines suppression for other fields than docs.defaults

### LINKED ISSUE #642: Multi-lines formatting issue in markdown
Jsii-docgen generates API documentation with only 2 lines: the first sentence and the rest of the code comment.

It generates incorrect output in various cases including:
- Only the first comment line is interpreted with a CR. The other CRs are not interpreted
- Code block is one single line, no indentation
- bullet lists are not formatted properly but all items are put into the same line with "*"

Example here :
- The code comment https://github.com/aws-samples/aws-analytics-reference-architecture/blob/d9f06ff7127754368ee1f6a9ffa472750c04c7d4/core/src/data-lake-storage.ts#L50
- The generated API.md https://github.com/aws-samples/aws-analytics-reference-architecture/blob/main/core/API.md#datalakestorage-

vgkowski:
Fixes #642 

The newlines should be removed only from docs.defaults
This commit isolates the newlines removal in a dedicated method which is only used on docs.defaults
Other fields continue to use sanitize() but without any newline removal

### DIFF
--- a/src/docgen/render/markdown-doc.ts
+++ b/src/docgen/render/markdown-doc.ts
@@ -70,9 +70,14 @@ export class MarkdownDocument {
       sanitized = sanitized.substring(1, line.length).trim();
     }
 
-    sanitized = sanitized.replace(/\n/g, ' ');
-
     return sanitized;
+  }
+
+  /**
+   * Remove newlines from markdown.
+   */
+  public static removeNewlines(line: string): string {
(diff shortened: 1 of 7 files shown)


Chriscbr (changes requested):
Please remove the changes to `yarn.lock` and `package-lock.json`. Otherwise LGTM

Chriscbr (approved):

PR #644, opened 2022-03-22
fix: support hoisted jsii dependencies

### LINKED ISSUE #390: Bug: Issue with temporary folder and symlinks
This issue occurs with the following project structure:

```
- root
  - packages
    - jsii project a
      - node_modules/project b -> symlink to project b containing '../../jsii project b'
    - jsii project b
```

jsii-reflect will correctly resolve all types. But jsii-docgen throws:
"Error: Assembly "jsii project b" not found"

This is (probably) due to the fact that the below code is making a full copy of the project structure:

https://github.com/cdklabs/jsii-docgen/blob/f3f0509f114e03274854c4f1d81ac5cf7691730c/src/docgen/view/documentation.ts#L180

The above code will copy the symlink as file to the temporary folder. This symlink will be invalid if its a relative link.

Probably either needs to fix it by not making a copy or to use something like:
```
await fs.copy(assembliesDir, workdir, {
    dereference: true
});
```

The copy also is responsible for a very long execution time due to all dependency copies. Didn't go through all code in the docgen package but it it really needed in this scenario to make a copy? I would think that the generator will not modify any files. Only the generation of the md file. Not copying would also make it possible to work without nohoist for peer dependencies.

This is actually part of a yarn monorepo where I configured the projects with nohoist. I cannot imagine i'm the only one encountering this. Anyone knows a way around it?

MrArnoldPalmer:
Removes copy of the project directory and instead only outputs
transliterated assemblies to the working directory. This avoids
potential problems when generating docs for packages that may have some
dependencies that are symlinks, such as those in a monorepo.

Pending https://github.com/aws/jsii/pull/3437

fixes #390


MrArnoldPalmer:
NOTE: Originally this was targeting allowing users to transliterate assemblies for projects that may have hoisted dependencies, IE a dependency is within a parent directory's `node_modules` somewhere. However, just changing to using the package directory as the working directory and outputting the transliterated assemblies elsewhere isn't good enough, as the `assembliesDir` defaults to the project directory and wont search parent directories for dependencies/.jsii assemblies.

This however should still fix the resolution of symlinked dependencies that may exist in a packages `node_modules` such as those within a lerna managed monorepo. This case is slightly harder to setup a succint test for but open to suggestions if reviewers feel strongly about adding one.

Chriscbr (approved):

agdimech:
This PR also improves performance by ~ 6x :)

One of my builds is down from 6 mins to 1 min with this PR.

MrArnoldPalmer:
yeah tests are fine for me as well but I haven't bothered trying to fix the flakiness in a different PR yet.

agdimech:
@Chriscbr - are we able to merge this?

agdimech:
Couldn't figure out how to contribute to this PR so I opened a fresh one which should fix the unit tests: https://github.com/cdklabs/jsii-docgen/pull/714

MrArnoldPalmer:
@agdimech ty! I didn't realize it was purely a timeout as I tried previously upping it and still seeing failures. I merged in your commit here just to prevent re-review etc. Should auto-merge when successful.

agdimech:
Sorry, I guess that didn’t fix it :(

I’ve seen this combination of errors locally to which I bumped the timeout and it resolved. Guess my hardware might be a bit beefier then the hw used as part of the build.

We can try bumping this to something like 120 although there may be a nasty timing issue contributing to this. What is strange is I would have thought that the test duration would have reduced with this PR, given the lack of fs.copy.

agdimech:
I just had a look at the other open PR: https://github.com/cdklabs/jsii-docgen/runs/6928391566?check_suite_focus=true

the same issue is showing up there related to: 
`ENOENT: no such file or directory, chdir '/tmp/EtnTZW' -> '/tmp/wryVeX `. What is really strange is it is failing when trying to cd back the the previous cwd as per: https://github.com/cdklabs/jsii-docgen/blob/dd39bac639905ff4166855f22fd0f60e6edf8b13/src/docgen/view/documentation.ts#L382

RichiCoder1:
I'm manually applying this patch to a local installation and I hope it moves forward if only because the _massive_ perf boost.

agdimech:
And thoughts on my previous comments?

I don’t think these build failures are related to this PR as per: https://github.com/cdklabs/jsii-docgen/runs/7122708633?check_suite_focus=true

agdimech:
Any idea how to proceed with this?

RichiCoder1:
Plusing this again. Fix both perf issues and support for Yarn/PNP which hoist dependencies.

agdimech:
This can be closed as #981 has been merged and released as of version ^7.2.1 :)

MrArnoldPalmer:
closing in favor of https://github.com/cdklabs/jsii-docgen/pull/981

PR #651, opened 2022-04-07
fix: transliterate snippets missing from tablet

MrArnoldPalmer:
Catch failures of `transliterateAssembly` and rethrow as
`LanguageNotSupportedError`. This allows adding support for Go as some
package versions were released with Go support before jsii-rosetta
supported it and have a tablet file without Go snippets.

In the future live transliteration of these missing snippets may be
ideal but the only package I've found with this scenario is aws-cdk-lib,
which takes a really long time to live transliterate and is something we
may want to optimize further before enabling.

Furthermore only a handful of aws-cdk-lib versions actually have this
problem as the latest have correct Go snippets in the tablet, andd the
earliest dont' have tablet files at all and are therefore fully transliterated
by jsii-docgen.


API.md
```diff
@@ -0,0 +1,754 @@
+# API Reference <a name="API Reference" id="api-reference"></a>
```
Chriscbr, line 1: Remove this file

Chriscbr (changes requested):

Chriscbr (approved):
Approving with `do-not-merge` label so you can merge it in after fixing the nit 👍

src/docgen/view/documentation.ts
```diff
@@ -328,7 +329,11 @@ export class Documentation {
         const spec = JSON.parse(await fs.readFile(dotJsii, 'utf-8'));
         if (language && spec.name === this.assemblyName) {
           const packageDir = path.dirname(dotJsii);
-          await transliterateAssembly([packageDir], [language], { loose: options.loose });
+          try {
+            await transliterateAssembly([packageDir], [language], { loose: options.loose, unknownSnippets: UnknownSnippetMode.FAIL });
+          } catch (e) {
+            throw new LanguageNotSupportedError(`Laguage ${language} is not supported for package ${this.assemblyFqn}`);
```
RomainMuller, line 335: Could have mentioned that this is a snippet translation issue....

PR #714, opened 2022-06-17
fix: support hoisted jsii dependencies

### LINKED ISSUE #390: Bug: Issue with temporary folder and symlinks

agdimech:
Removes copy of the project directory and instead only outputs
transliterated assemblies to the working directory. This avoids
potential problems when generating docs for packages that may have some
dependencies hoisted. IE when building against `aws-cdk-lib` within
`node_modules` and `constructs` is hoisted to the top level
`node_modules` directory as well.

Note: This fixes unit tests for https://github.com/cdklabs/jsii-docgen/pull/644. Full credit goes to Mitchell Valine @MrArnoldPalmer.

Fixes #390

### DIFF
--- a/src/docgen/view/documentation.ts
+++ b/src/docgen/view/documentation.ts
@@ -316,11 +316,8 @@ export class Documentation {
 
     const created = await withTempDir(async (workdir: string) => {
 
-      // always better not to pollute an externally provided directory
-      await fs.copy(this.assembliesDir, workdir);
-
       const ts = new reflect.TypeSystem();
-      for (let dotJsii of await glob.promise(`${workdir}/**/.jsii`)) {
+      for (let dotJsii of await glob.promise(`${this.assembliesDir}/**/.jsii`)) {
         // we only transliterate the top level assembly and not the entire type-system.
         // note that the only reason to translate dependant assemblies is to show code examples
         // for expanded python arguments - which we don't to right now anyway.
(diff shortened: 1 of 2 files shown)


agdimech:
Note: This fixes unit tests for https://github.com/cdklabs/jsii-docgen/pull/644. Full credit goes to Mitchell Valine @MrArnoldPalmer.

Happy to either merge this in or have Mitchell apply my unit test fix to his original PR 😄

agdimech:
Closing as this was merged in to #644

PR #934, opened 2023-02-20
fix: remove double .md suffix

DanielMSchmidt:


src/cli.ts
```diff
@@ -18,9 +18,10 @@ async function generateForLanguage(docs: Documentation, options: GenerateOptions
   const fileSuffix = format === 'md' ? 'md' : 'json';
   let submoduleSuffix = fileSuffix;
 
+  const outputFileName = output.replace(/\.md|\.json/g, '');
```
RomainMuller, line 21: Should we only strip the suffix if it's the expected one?

```suggestion
  const outputFileName = output.endsWith(`.${fileSuffix}`)
    ? output.slice(0, -(fileSuffix.length + 1))
    : output;
```

RomainMuller (approved):

PR #1037, opened 2023-06-09
fix: optional peer dependencies are not installed in dependency closure

mrgrain:
Peer dependencies can be marked as optional via `peerDependenciesMeta`.
This prevents them from being installed by npm and causes jsii-docgen to fail because of a missing assembly.
Example: https://www.npmjs.com/package/@lumigo/cdk-constructs-v2

With this change we explicitly install optional peer dependencies.


RomainMuller (approved):

mrgrain:
Drafting for now. The proposed fix fails for "local" (*) packages. 
Need to adjust the approach for this case to detect optional peer dependencies differently or maybe just bypass the check.

(*) Anything that's not on npmjs.com, like a tarball or a folder.

src/docgen/view/_npm.ts
```diff
@@ -19,21 +21,28 @@ export class Npm {
...
+    assertSuccess(await this.runCommand(
       await this.npmCommandPath(),
       [
         'install',
         target,
         // this is critical from a security perspective to prevent
         // code execution as part of the install command using npm hooks. (e.g postInstall)
         '--ignore-scripts',
+        // save time by not running audit
+        '--no-autit',
```
mrgrain, line 37:
```suggestion
        '--no-audit',
```

src/docgen/view/_npm.ts
```diff
@@ -44,8 +53,62 @@ export class Npm {
...
+      assertSuccess(await this.runCommand(
+        await this.npmCommandPath(),
+        [
+          'install',
+          ...optionalPeerDeps,
+          // this is critical from a security perspective to prevent
+          // code execution as part of the install command using npm hooks. (e.g postInstall)
+          '--ignore-scripts',
+          // save time by not running audit
+          '--no-autit',
```
mrgrain, line 75:
```suggestion
          '--no-audit',
```

PR #1042, opened 2023-06-13
feat!: upgrade jsii-rosetta to 5.1.x and make it a peer dependency

mrgrain:
BREAKING CHANGE: Change `jsii-rosetta` to be a peer dependency with a tested version range. Minimum version is `5.1.2` with this release and we will add new minor versions once they are tested.

Package managers will attempt to install the peer dependency, unless explicitly configured not to.
You can install newer versions of `jsii-rosetta` when they are released, but they should be considered untested until the peer dependency version range is updated.


RomainMuller (changes requested):

rix0rrr (approved):

.github/workflows/build.yml
```diff
@@ -76,3 +76,33 @@ jobs:
...
+          ref: ${{ github.event.pull_request.head.ref }}
+          repository: ${{ github.event.pull_request.head.repo.full_name }}
+      - name: Setup Node.js
+        uses: actions/setup-node@v3
+        with:
+          node-version: 16.x
+      - name: Install dependencies
+        run: yarn install --check-files
+      - name: Install Rosetta version
+        run: yarn add --dev jsii-rosetta@${{ matrix.rosetta }}
```
RomainMuller, line 98: Wonder if we can slap a `--no-save` here instead to avoid changing the `package.json` as a side-effect.

```suggestion
        run: yarn add --no-save jsii-rosetta@${{ matrix.rosetta }}
```

.github/workflows/build.yml
```diff
@@ -76,3 +76,33 @@ jobs:
...
+      - name: Check Rosetta version
+        run: test $(npx jsii-rosetta --version) = "${{ matrix.rosetta }}"
+      - name: compile+test
+        run: |-
+          npx projen compile
+          npx projen test
```
RomainMuller, line 104: I suspect running `npx projen ...` will run `yarn install` which may cause the `jsii-rosetta` to be pruned or reverted... As a result, I'd actually rather have the `jsii-rosetta --version` test done **after** this as well...

Also - this doesn't quite test that everything works "as intended" when it was built with `jsii-rosetta@5.1.x` but is running with `jsii-rosetta@5.2.x` (hypothetically).

mrgrain: projen doesn't run an install here, but I sure can move the check. 👍

.github/workflows/build.yml
mrgrain: `yarn` doesn't have that feature unfortunately.

.github/workflows/build.yml
mrgrain: > Also - this doesn't quite test that everything works "as intended" when it was built with jsii-rosetta@5.1.x but is running with jsii-rosetta@5.2.x (hypothetically).

Yes ideally this would take the same approach as we e.g. do with `aws-cdk-lib`. Where we'd build the package to a tarball. Then put that tarball + desired rosetta version in an empty workspace and then run an integration test suite.
But the package isn't structured that way (yet) so that would be a lot more work. **Is it worth it?**
'Cause we we don't really build this package "with" `jsii-rosetta`. We run `tsc` and that's about it. Not bundling or whatsoever. Technically the current test suite doesn't even test the artifact, as it's running with `ts-node`.

RomainMuller (approved):

PR #1117, opened 2023-08-25
fix!: readme incorrectly included in API docs by default

### LINKED ISSUE #1073: --readme option documentation and implementation do not align
when running the CLI without the `-r` flag, it defaults to true but this shouldn't be the case according to the documentation.

https://github.com/cdklabs/jsii-docgen/blob/a159e5fd7bf6540af206175acd823eb8762324d2/src/docgen/view/documentation.ts#L251-L253

Are you accepting contributing for this?

oielbanna:
Fixes #1073 

BREAKING CHANGE: README.md files are no longer included by default in the generated documentation files. To revert to the previous behavior, please provide the `--readme` flag.


mrgrain (approved):

mrgrain:
Hey @oielbanna Thanks for this! I've updated the PR description etc to properly mark this as a breaking change. Can you update the test snapshots to make the build pass?

mrgrain:
> > Hey @oielbanna Thanks for this! I've updated the PR description etc to properly mark this as a breaking change. Can you update the test snapshots to make the build pass?
> 
> Sorry did not see your comment until now. I see you've updated the snapshot already so thanks for doing that!

No worries. I'm on it now. It's been a pain, cause there's a test that is explicitly skipped on MacOS 😢

PR #1190, opened 2023-11-02
feat: Documentation.toJson() can handle nested submodules

otaviomacedo:
Allow users to create documentation for nested modules. Suppose a project has the following submodule structure:

```
my-project
└── foo
    └── bar
```

where `foo` and `bar` are submodules. To generate documentation for the `bar` module, call `documentation.toJson('foo.bar')`.


src/docgen/view/documentation.ts
```diff
@@ -315,19 +315,25 @@ export class Documentation {
...
-    if (submodules.length > 1) {
-      throw new Error(`Found multiple submodules with name: ${submodule} in assembly ${assembly.name}@${assembly.version}`);
-    }
+      if (found.length === 0) {
+        throw new Error(`Submodule ${submodule} not found in assembly ${assembly.name}@${assembly.version}`);
+      }
 
-    return submodules[0];
+      if (found.length > 1) {
+        throw new Error(`Found multiple submodules with name: ${submodule} in assembly ${assembly.name}@${assembly.version}`);
```
rix0rrr, line 332: I'm pretty sure this can never happen, but okay.

src/docgen/view/documentation.ts
```diff
@@ -315,19 +315,25 @@ export class Documentation {
    * Lookup a submodule by a submodule name.
```
rix0rrr, line 315: Maybe update the docs to indicate what we now support?

otaviomacedo: Done.

rix0rrr (approved):

PR #1251, opened 2023-12-20
fix: cannot use splitBySubmodule and multiple languages together

### LINKED ISSUE #1250: --split-by-submodule

mrgrain:
Fixes #1250

Given the command

```console
jsii-docgen -o docs/api/API.md --split-by-submodule -l typescript -l python -l java -l go
```

the output now looks like this:

image

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -10,26 +10,18 @@ type GenerateOptions = {
   submodule?: string;
   allSubmodules?: boolean;
   splitBySubmodules?: boolean;
-  format?: 'md' | 'json';
-  output?: string;
+  format: 'md' | 'json';
+  output: string;
 }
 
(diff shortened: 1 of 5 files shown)


rix0rrr:
Will this work fine in Construct Hub?

rix0rrr:
How are the bases determined? `API` vs `awscdk` vs `build` ?

mrgrain:
> Will this work fine in Construct Hub?

ConstructHub doesn't use the CLI, so this won't have any impact.

> How are the bases determined? API vs awscdk vs build ?

The bases are determined exactly the same way as before, this hasn't changed in this PR. `API` is whatever is passed in as `--output`. `awscdk` etc. are the names of the submodules.

rix0rrr (approved):

PR #1375, opened 2024-04-08
feat: support `jsii-rosetta` 5.4 version line

iliapolo:
Following https://github.com/aws/jsii-rosetta/pull/1110


.projenrc.ts
```diff
@@ -61,13 +61,14 @@ const project = new CdklabsTypeScriptProject({
 // Suppress upgrade prompts, in particular from test runs.
 project.tasks.addEnvironment('JSII_SUPPRESS_UPGRADE_PROMPT', 'true');
 
+// Add the new version line here, and remove any that have reached EOS.
 new RosettaPeerDependency(project, {
   supportedVersions: {
     [RosettaVersionLines.V1_X]: '^1.85.0',
-    [RosettaVersionLines.V5_0]: '~5.0.14',
```
mrgrain, line 67: Removing 5.0 would be a breaking change since `jsii-docgen` is following plain semver (contrary to `jsii-rosetta`).
Feels like no harm just keeping it (at least for now)?! Otherwise we should mark the PR as breaking.

mrgrain (changes requested):

.projenrc.ts
iliapolo: Hmm, yeah ok, lets keep it.

mrgrain (approved):

PR #1398, opened 2024-04-28
fix: Assembly not found on Windows

mrgrain:
jsii-docgen fails in Windows with the error `Assembly "<assembly-name>" not found`. 
The reason for this is that the path passed to the glob library we use must only use forward slashes.
However on Windows when using jsii-docgen, it may contain backslashes. 

To fix the issue, we pass the assemblies directory as CWD to glob instead. The CWD doesn't have the same limitation in regards to backslashes.
Other fixes involve making the npm helpers Windows aware.

And finally this PR adds a Windows test to the compat build matrix.


src/docgen/view/_npm.ts
```diff
@@ -38,7 +38,7 @@ export class Npm {
       // code execution as part of the install command using npm hooks. (e.g postInstall)
       '--ignore-scripts',
       // save time by not running audit
-      '--no-autit',
+      '--no-audit',
```
mrgrain, line 41: 🙈

rix0rrr (approved):

PR #1556, opened 2024-09-03
fix: variadic arguments are incorrectly displayed as base type in docs

### LINKED ISSUE #648: Variadic arguments are not treated as such
For example, a method signature like so:

```ts
public add(...scopes: IDependable[])
```

Translates into:

```java
public void add(IDependable scopes)
```

This is true for all languages. 

> See https://constructs.dev/packages/constructs/v/10.0.98/api/DependencyGroup?lang=java for the example.

sumupitchayan:
Fixes #648 

Fixes issue where variadic arguments are incorrectly displayed as the base type in documentation.

Example from the original issue:

```ts
public add(...scopes: IDependable[])
```
is shown as:

```ts
public void add(IDependable scopes)
```


src/docgen/render/markdown-render.ts
```diff
@@ -630,7 +630,9 @@ export class MarkdownRenderer {
     }
 
     for (const [key, value] of Object.entries(metadata)) {
-      md.bullet(`${MarkdownDocument.italic(`${key}:`)} ${value}`);
+      // To denote if a parameter is variadic
+      const variadicMarker = parameter.variadic ? 'variadic ' : '';
+      md.bullet(`${MarkdownDocument.italic(`${key}:`)} ${variadicMarker}${value}`);
```
sumupitchayan, line 635: Not sure if this is exactly how we want to do this - right now I set this so docs show a parameter is variadic like so:

**Type:** variadic string

I could change it to be each language's specific way for declaring a variadic parameter

iliapolo: Yes this should be language specific. Might be a little tricky. lets talk.

test/docgen/assemblies.ts
```diff
@@ -29,6 +29,11 @@ export class Assemblies {
     this.addAssemblies(Assemblies.AWSCDK_1_106_0);
   }
 
+  // To test a simple assembly with variadic arguments
+  public get variadicExampleTest(): reflect.Assembly {
```
iliapolo, line 33: For consistent naming:

```suggestion
  public get withVariadicParameter(): reflect.Assembly {
```

test/docgen/view/initializer.test.ts
```diff
@@ -28,3 +28,42 @@ test.each(Language.values())('%s snapshot', (language) => {
   expect(init).toMatchSnapshot();
   expect(markdown.visitInitializer(init).render()).toMatchSnapshot();
 });
+
+const variadicTestAssembly: reflect.Assembly = Assemblies.instance.variadicExampleTest;
+const findVariadicInitializer = (): reflect.Initializer => {
+  for (const klass of variadicTestAssembly.classes) {
+    if (klass.initializer) {
```
iliapolo, line 35: You are assuming here that the first initializer in the assembly has a variadic parameter. Its true now, but will it always be? 

This function needs to search for exactly what it needs, and not make any assumptions on how the assembly looks like.

test/docgen/view/initializer.test.ts
```diff
@@ -28,3 +28,42 @@ test.each(Language.values())('%s snapshot', (language) => {
...
+      return klass.initializer;
+    }
+  }
+  throw new Error('Assembly does not contain an initializer');
```
iliapolo, line 39: Adjust error based on what the function looks for.

test/docgen/view/initializer.test.ts
```diff
@@ -28,3 +28,42 @@ test.each(Language.values())('%s snapshot', (language) => {
...
+        if (param.variadic) {
+          return param;
+        }
+      }
+    }
+  }
+  throw new Error('Assembly does not contain a parameter');
+};
+
+test('variadic parameter in initializer is correct', () => {
```
iliapolo, line 55: You're testing here that a test function behaves like you expect. Tests shouldn't be testing tests :)

If the `findVariadicParameter` test is incorrect, than the snapshot would be incorrect, and you should iterate until it is.

test/docgen/view/initializer.test.ts
```diff
@@ -28,3 +28,42 @@ test.each(Language.values())('%s snapshot', (language) => {
...
+  expect(variadicParameter.variadic).toEqual(true);
+});
+
+test.each(Language.values())('%s snapshot', (language) => {
+  const { transpile } = LANGUAGE_SPECIFIC[language.toString()];
+  const renderer = new MarkdownRenderer({ language, ...metadata });
+  const init = new Initializer(transpile, findVariadicInitializer()).toJson();
+  const markdown = renderer.visitInitializer(init).render();
+  // This asserts that the type parameter is marked as variadic:
+  expect(markdown).toContain('*Type:* variadic');
```
iliapolo, line 67: No need for this given you're testing the entire snapshot

test/docgen/view/initializer.test.ts
```diff
@@ -28,3 +28,42 @@ test.each(Language.values())('%s snapshot', (language) => {
...
+  for (const klass of variadicTestAssembly.classes) {
+    if (klass.initializer) {
+      for (const param of klass.initializer.parameters) {
+        if (param.variadic) {
+          return param;
+        }
+      }
+    }
+  }
+  throw new Error('Assembly does not contain a parameter');
```
iliapolo, line 52:
```suggestion
  throw new Error('Assembly does not contain a variadic parameter');
```

test/docgen/view/initializer.test.ts
```diff
@@ -28,3 +28,42 @@ test.each(Language.values())('%s snapshot', (language) => {
...
+test('variadic parameter in initializer is correct', () => {
+  const variadicParameter = findVariadicParameter();
+  expect(variadicParameter.name).toEqual('variadicProps');
+  expect(variadicParameter.variadic).toEqual(true);
+});
+
+test.each(Language.values())('%s snapshot', (language) => {
+  const { transpile } = LANGUAGE_SPECIFIC[language.toString()];
+  const renderer = new MarkdownRenderer({ language, ...metadata });
+  const init = new Initializer(transpile, findVariadicInitializer()).toJson();
```
iliapolo, line 64: Shouldn't this use `findVariadicParameter` and moved to `parameter.test.ts`? This would then cover all usages of parameter (e.g initializer, instance-method, and static function).

src/docgen/render/markdown-render.ts
iliapolo, line 635: Even if we would do it in a language agnostic way, this is not the right for it. think why.

iliapolo (changes requested):
Couple high level comments:

- We talked about fixing the `assembly.system.classes` to use `assembly.classes` in the various lookup functions. Are you planning on doing this in a separate PR? 
- PR title should indicate what the specific problem was. Remember that it is the first thing customers see in the change log, so it should be as specific as possible. How exactly were variadics wrong in the docs?

sumupitchayan:
> * We talked about fixing the `assembly.system.classes` to use `assembly.classes` in the various lookup functions. Are you planning on doing this in a separate PR?

Yea, I think it'd be better to separate that from this fix. This PR would be cluttered with the snapshot updates from both fixes

test/docgen/view/initializer.test.ts
sumupitchayan: Are you saying to remove the test from `initializer.test.ts` and just add this one in `parameter.test.ts`? Makes sense to me. Would we also remove the test in `instance-method.test.ts` either, since the `parameter` test applies everywhere?

iliapolo: Yes

src/docgen/view/parameter.ts
```diff
@@ -24,6 +24,7 @@ export class Parameter {
       optional: this.transpiledParam.optional === true ? true : undefined, // to save space
       default: this.parameter.spec.docs?.default,
       type: this.transpiledParam.typeReference.toJson(),
+      variadic: this.parameter.spec.variadic ?? false,
```
iliapolo, line 27:
```suggestion
      variadic: this.transpiledParam.variadic,
```

src/docgen/transpile/python.ts
```diff
@@ -150,6 +150,7 @@ export class PythonTranspile extends transpile.TranspileBase {
       parentType: this.type(property.parentType),
       typeReference: typeRef,
       optional: property.optional,
+      variadic: false,
```
iliapolo, line 153: This seems like a good reason to create a distinction between the `TranspiledProperty` and `TranspiledParam` interfaces. See https://www.typescriptlang.org/docs/handbook/utility-types.html#omittype-keys

src/docgen/transpile/typescript.ts
```diff
@@ -236,6 +238,11 @@ export class TypeScriptTranspile extends transpile.TranspileBase {
     const tf = transpiled.typeReference.toString({
       typeFormatter: (t) => t.name,
     });
+
+    if (transpiled.variadic) {
+      return `${transpiled.name}${transpiled.optional ? '?' : ''}: ...${tf}[]`;
```
iliapolo, line 243:
```suggestion
      return `${transpiled.name}${transpiled.optional ? '?' : ''}: ...${this.listOf(tf)}`;
```

Think about whether this is reusable enough. As you mentioned, this PR is still missing the variadic formatting in the `Type` section of the metadata. This probably means you'll have to repeat this logic somewhere else - how can we make sure we only define this formatting once?

test/docgen/assemblies.ts
```diff
@@ -29,6 +29,11 @@ export class Assemblies {
     this.addAssemblies(Assemblies.AWSCDK_1_106_0);
   }
 
+  // To test a simple assembly with variadic arguments
+  public get withVariadicParameter(): reflect.Assembly {
+    return this.ts.findAssembly('my-construct-library');
```
iliapolo, line 34: Give your assembly a more indicative name than `my-construct-library`

src/docgen/view/parameter.ts
iliapolo: At a high level, the way docgen works is:

`jsii.spec -> jsii.reflect -> transpiled -> schema -> markdown`

As we discussed, the markdown rendering should be language agnostic, where the schema contains every language specific formatting. Is adding `variadic` to the schema enough for the renderer to know how to format types? its ok to stick it in here as possible useful metadata, but its not enough (this is why metadata is still incorrect).

test/docgen/view/parameter.test.ts
```diff
@@ -52,3 +53,71 @@ test('newlines in any other than "defaults" are not removed', () => {
...
+        if (param.variadic) {
+          return param;
+        }
+      }
+    }
+  }
+  throw new Error('Assembly does not contain a variadic parameter');
+};
+
+test.each(Language.values())('%s snapshot', (language) => {
```
iliapolo, line 71:
```suggestion
test.each(Language.values())('%s snapshot variadic', (language) => {
```

etc...

iliapolo (changes requested):

src/docgen/transpile/python.ts
```diff
@@ -306,6 +308,11 @@ export class PythonTranspile extends transpile.TranspileBase {
     const tf = transpiled.typeReference.toString({
       typeFormatter: (t) => t.name,
     });
+
+    if (transpiled.variadic) {
+      return `${transpiled.name}: *${tf}${transpiled.optional ? ' = None' : ''}`;
```
iliapolo, line 313: Can variadic parameters be optional?

sumupitchayan: Nope! fixed

src/docgen/transpile/csharp.ts
```diff
@@ -103,7 +103,7 @@ export class CSharpTranspile extends transpile.TranspileBase {
     const type = this.type(struct);
     const indent = ' '.repeat(4);
     const inputs = struct.allProperties.map((p) => {
-      return `${indent}${this.formatFnParam(this.property(p))}`;
+      return `${indent}${this.formatFnParam(this.parameter(p))}`;
```
iliapolo, line 106: Can you explain? `struct.allProperties` returns `reflect.Property` - so why do we have to call `this.parameter` instead of `this.property`?

src/docgen/transpile/typescript.ts
```diff
@@ -70,6 +70,10 @@ export class TypeScriptTranspile extends transpile.TranspileBase {
     return `${type}[]`;
   }
 
+  public variadicOf(type: string): string {
+    return `...${type}[]`;
```
iliapolo, line 74:
```suggestion
    return `...${this.listOf(type)}`;
```

src/docgen/view/parameter.ts
```diff
@@ -6,14 +6,21 @@ export class Parameter {
...
+    private readonly transpile: Transpile,
     private readonly parameter: reflect.Parameter,
   ) {
     this.transpiledParam = transpile.parameter(parameter);
     this.transpiledCallable = transpile.callable(parameter.method);
   }
 
   public toJson(): ParameterSchema {
+    let typeschema = this.transpiledParam.typeReference.toJson();
+    if (this.parameter.spec.variadic) {
```
iliapolo, line 18:
```suggestion
    if (this.transpiledParam.variadic) {
```

src/docgen/view/parameter.ts
```diff
@@ -6,14 +6,21 @@ export class Parameter {
...
+      typeschema = {
+        formattingPattern: this.transpile.variadicOf(this.transpiledParam.typeReference.toString()),
```
iliapolo, line 20: This is tricky to understand. Please explain with some examples.

src/docgen/transpile/csharp.ts
```diff
@@ -170,6 +171,10 @@ export class CSharpTranspile extends transpile.TranspileBase {
     return `${type}[]`;
   }
 
+  public variadicOf(type: string): string {
+    return `params ${type}[]`;
```
iliapolo, line 175:
```suggestion
    return `params ${this.listOf(type)}`;
```

test/docgen/view/__snapshots__/markdown.test.ts.snap
```diff
@@ -6473,18 +6473,18 @@ construct tree needs to be combined to be used as a single dependable.
...
 | **Name** | **Type** | **Description** |
 | --- | --- | --- |
-| <code><a href="/packages/constructs/v/10.0.78/api/DependencyGroup.Initializer.parameter.deps?lang=python">deps</a></code> | <code><a href="/packages/constructs/v/99.99.99/api/IDependable?lang=python">IDependable</a></code> | *No description.* |
+| <code><a href="/packages/constructs/v/10.0.78/api/DependencyGroup.Initializer.parameter.deps?lang=python">deps</a></code> | <code>...github.com/aws/constructs-go/constructs/v10.IDependable</code> | *No description.* |
 
 ---
 
 ##### \`deps\`<sup>Required</sup> <a name="deps" id="DependencyGroup.Initializer.parameter.deps"></a>
 
-- *Type:* <a href="/packages/constructs/v/99.99.99/api/IDependable?lang=python">IDependable</a>
```
iliapolo, line 6487: Hmm, this doesn't look right...somehow we are dropping the link. This should be:

```console
...<a href="/packages/constructs/v/99.99.99/api/IDependable?lang=python">IDependable</a>
```

src/docgen/transpile/csharp.ts
sumupitchayan: My mistake here, this is unnecessary - reverted to the original `this.property(p)`

src/docgen/transpile/csharp.ts
```diff
@@ -123,7 +123,7 @@ export class CSharpTranspile extends transpile.TranspileBase {
   }
 
   public parameter(
-    parameter: reflect.Parameter,
+    parameter: reflect.Parameter | reflect.Property,
```
iliapolo, line 126:
```suggestion
    parameter: reflect.Parameter,
```

src/docgen/transpile/csharp.ts
```diff
@@ -132,6 +132,7 @@ export class CSharpTranspile extends transpile.TranspileBase {
       parentType: this.type(parameter.parentType),
       typeReference: typeRef,
       optional: parameter.optional,
+      variadic: 'variadic' in parameter ? parameter.variadic : false,
```
iliapolo, line 135:
```suggestion
      variadic: parameter.variadic,
```

iliapolo (approved):

iliapolo:
@sumupitchayan Any reason this PR is still in draft?

PR #1896, opened 2026-03-16
fix: removing unnecessary copy

TheCynosure:
## Why

This PR fixes #1895.

From researching the code-base, the only explicit output we do is when we render the final documentation. The actual type system compilation isn't producing any "polluting" files so this copy is unnecessary. 

This copy is causing a massive slowdown in our project repository because our `node_modules` is quite large (nearing ~1Gb).

## Testing

I tested this on our code-base and it doesn't produce any extra files in the node_modules or the working directory besides the output documentation.

### DIFF
--- a/src/docgen/view/documentation.ts
+++ b/src/docgen/view/documentation.ts
@@ -405,16 +405,6 @@ export class Documentation {
 
     const created = await withTempDir(async (workdir: string) => {
 
-      // always better not to pollute an externally provided directory
-      await fs.copy(this.assembliesDir, workdir, {
-        // Ensure we don't try to copy socket files, as they can be found under .git when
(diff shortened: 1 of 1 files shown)


mrgrain (changes requested):
This might fail for language targets, because later on we try to load the assembly from the workdir:

https://github.com/TheCynosure/jsii-docgen/blob/7a23b52479bd40ecf2f2c8906d338e1d94cd8786/src/docgen/view/documentation.ts#L427

So I guess if we update `loadAssembly` to somehow use the original `this.assembliesDir` as lookup base, we can remove the copy.

Nice improvement!

automartin5000:
thanks so much @mrgrain and @iliapolo!

